PR #161, opened 2014-11-26
Track instances

### LINKED ISSUE #140: Add verbosity to Execute actor
When executing on multiple instances there is no insight into which tasks have completed and which ones are blocking. Add "Done" log, as well as minutely report saying "Still waiting on ..." similar to how we do in Launch actor.

CC: @chuckkarish @diranged

siminm:
Close #140

### DIFF
--- a/kingpin/actors/rightscale/api.py
+++ b/kingpin/actors/rightscale/api.py
@@ -409,11 +409,7 @@ class RightScale(object):
 
         return task
 
-    @concurrent.run_on_executor
-    @sync_retry(stop_max_attempt_number=20,
-                wait_exponential_multiplier=1000,
-                wait_exponential_max=10000)
-    @utils.exception_logger
+    @gen.coroutine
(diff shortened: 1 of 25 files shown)


kingpin/actors/rightscale/api.py
```diff
@@ -513,18 +532,28 @@ def run_executable_on_instances(self, name, inputs, instances):
...
 
-        raise gen.Return(ret)
+        # At this point all the tasks are executed, but the reference
+        # that we store in task_pairs are still to the Future objects.
+        yielded_tasks = []
+        for (i, task) in task_pairs:
+            result = yield task  # Should be a no-op since it's completed above
+            yielded_tasks.append((i, result))
+
+        raise gen.Return(yielded_tasks)
```
diranged: You've changed the return value here, so the docstrings above are wrong. Please update them.

diranged:
Clean up the 1 docstring, and then it looks good. :+1:

PR #168, opened 2015-01-01
Use REQUIRED constant instead of None for options

siminm:
Sometimes None is a valid default value, and it should not be reserved
to identify a required parameter. This change converts all actors to use
a REQUIRED identifier (meta class).

### DIFF
--- a/kingpin/actors/base.py
+++ b/kingpin/actors/base.py
@@ -39,6 +39,7 @@ from tornado import httputil
 
 from kingpin import utils
 from kingpin.actors import exceptions
+from kingpin.constants import REQUIRED
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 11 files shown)


kingpin/actors/base.py
```diff
@@ -132,10 +132,17 @@ def _validate_options(self):
         """
 
         # Loop through all_options, and find the required ones
-        # Required options have `None` as their default value.
+        self.log.debug(self.all_options.items())
+        if len(self.all_options.items()):
```
diranged: Whats going on here? The code is simple -- but I don't actually know what its doing.

siminm: Oh goodness. That wasn't supposed to make the commit.

siminm:
Removed the extra debug lines.

diranged:
:+1:

Issue #48, opened 2014-09-03
kingpin.actors.aws.cloudformation: Amazon AWS CloudFormation Actors


diranged:
We should support building complex AWS components using Cloudformation. The user should be able to define a template, create a stack, delete a stack and even update a stack. This should allow users to do far more than just what our built-in actors support, without a ton of work on our part.

  * kingpin.actors.aws.cloudformation.Create
  * kingpin.actors.aws.cloudformation.Delete
  * kingpin.actors.aws.cloudformation.Update

PR #169, opened 2015-01-05
Add actors to update ELB certs.

siminm:

PR #171, opened 2015-01-12
Add aws.cloudformation.Create/Delete actors to resolve issue #48

diranged:
Closes issue #48.

In the future we can add the Update actor, but updating CloudFormation templates is a relatively tricky thing.


kingpin/bin/deploy.py
```diff
@@ -93,8 +93,8 @@ def main():
                        options={'macro': options.json,
                                 'tokens': env_tokens})
         yield runner.execute()
-    except actor_exceptions.ActorException:
-        log.error('Kingpin encountered mistakes during the play.')
+    except actor_exceptions.ActorException as e:
+        log.error('Kingpin encountered mistakes during the play: %s' % e)
```
siminm: We're gonna have a merge-agreement: https://github.com/Nextdoor/kingpin/pull/169/files#diff-fb0822d0582573214d6359739e769803R98

diranged: Thats fine.. i don't really care which one wins either.

docs/actors/aws.cloudformation.Delete.md
```diff
@@ -0,0 +1,21 @@
...
+    { "desc": "Create production backend stack",
+      "actor": "aws.cloudformation.Create",
+      "options" {
+        "region": "us-west-1",
+        "name": "%CF_NAME%",
+      }
+    }
+
+**Dry Mode**
+
```
siminm: That's too dry. Let's at least do something

kingpin/actors/aws/cloudformation.py
```diff
@@ -0,0 +1,434 @@
...
+            sleep: (int) Time in seconds between stack status checks
+
+        Raises:
+            StackNotFound: If the stack doesn't exist.
+        """
+        while True:
+            stack = yield self._get_stack(self.option('name'))
+
+            if not stack:
+                msg = 'Stack %s not found?' % self.option('name')
```
siminm: This shouldn't be a question. Would also suggest wrapping %s as `"%s"` so that the message is never "Stack not found"

kingpin/actors/aws/cloudformation.py
```diff
@@ -0,0 +1,434 @@
...
+
+    """Raised on any generic CloudFormation error."""
+
+
+class InvalidTemplate(exceptions.UnrecoverableActorFailure):
+
+    """An invalid CloudFormation template was supplied."""
+
+
+class StackAlreadyExists(exceptions.RecoverableActorFailure):
```
siminm: This and NotFound should probably inherit from CloudFormationError, just to respect the hierarchy

kingpin/actors/aws/cloudformation.py
```diff
@@ -0,0 +1,434 @@
...
+            Tuple with:
+              (None/Contents of template file,
+               None/URL of template)
+
+        Raises:
+            InvalidTemplate
+        """
+        remote_types = ('http://', 'https://')
+
+        if self.option('template').startswith(remote_types):
```
siminm: I think you mean `template` here and not `self.option('template')`

kingpin/actors/aws/cloudformation.py
```diff
@@ -0,0 +1,434 @@
...
+            return (None, template)
+
+        try:
+            return (open(template, 'r').read(), None)
```
siminm: add `TODO: leverage self.readfile()` -- I'll do that after #169 is merged.

kingpin/actors/aws/cloudformation.py
```diff
@@ -0,0 +1,434 @@
...
+        If the template string supplied is a local file resource (has no
+        URI prefix), then this method will return the contents of the file.
+        Otherwise, returns None.
+
+        Args:
+            template: String with a reference to a template location.
+
+        Returns:
+            Tuple with:
+              (None/Contents of template file,
```
siminm: Returns `Contents/None` not `None/Contents` (or fix the code below)

kingpin/actors/aws/cloudformation.py
```diff
@@ -0,0 +1,434 @@
...
+                               'stack status becomes CREATE_FAILED'),
+    }
+
+    def __init__(self, *args, **kwargs):
+        """Initialize our object variables."""
+        super(Create, self).__init__(*args, **kwargs)
+
+        # Check if the supplied CF template is a local file. If it is, read it
+        # into memory.
+        (self._template_body, self._template_url) = self._get_template_body(
```
siminm: Not sure why this needs to be in the `__init__` and not just in `_execute()`.

Also add a comment here that this is for boto's `validate_template` otherwise it's confusing why you're preparing these.

kingpin/actors/aws/cloudformation.py
```diff
@@ -0,0 +1,434 @@
...
+    @utils.exception_logger
+    def _validate_template(self):
+        """Validates the CloudFormation template.
+
+        Raises:
+            InvalidTemplate
+            exceptions.InvalidCredentials
+        """
+        if self._template_body is not None:
+            self.log.debug('Validating template with AWS...')
```
siminm: I think these two should be `.info()` -- since it may take a second to check it would be valuable to see it.

siminm:
:+1:

PR #177, opened 2015-01-23
Print ELB health status every 30 seconds instead of 3

### LINKED ISSUE #150: Exponential Backoff for elb.WaitUntilHealthy Actor checks
This actor expects (correctly) that joining ELB is a fast action. Because of this the retries (and therefore log messages) are extremely frequent (3 seconds). This is good when everything works correctly, but if the joining fails, or hangs, the flood of messages becomes overwhelming.

I suggest changing the 3 seconds to 1, and every time it fails multiply the next wait time by 2.

siminm:
Close #150


diranged:
:+1:

PR #182, opened 2015-01-29
Handle 401 and 403 response codes the same.

diranged:
Rollbar has changed their response code on invalid_credentials
to respond with a 401 instead of a 403. Support both, just in case
they return to the old model.


diranged:
CC: @siminm

siminm:
Change to "in" syntax, then merge! :+1:

PR #183, opened 2015-01-30
actors.rightscale.server_array integration test fixes.

diranged:
1. Ensure that we never launch more than 1 instance, even if
   that instance takes forever. Effectively disables RS AutoScaling.
2. Fix an integration test that was supposed to break and throw/catch
   an exception, but wasn't because it was a bad copy-paste.


siminm:
:+1:

Issue #189, opened 2015-02-02
RightScale actors should observe OAUTH token timeout..


diranged:
The tokens expire about 2 hours after init.. Every response back includes a countdown till the token expires. We should honor this and regenerate a new token when necessary.

PR #190, opened 2015-02-02
Use the patched python-rightscale library with auto-token-refreshment.

diranged:
Pull request is pending with author:
https://github.com/brantai/python-rightscale/pull/11

(Issue #189)


diranged:
Verbally approved by @cmclaughlin..

PR #192, opened 2015-02-04
Initial kingpin.actors.support.api module for fast-creation of REST API ...

diranged:
This is a WIP ... will update the pull request as I work on the branch.

### DIFF
--- a/kingpin/actors/slack.py
+++ b/kingpin/actors/slack.py
@@ -123,7 +123,7 @@ class Message(SlackBase):
 
         # If we're in dry mode, bail out!
         if self._dry:
-            self.log.info('Skipping message send (dry run..)')
+            self.log.info('API Credentials verified, skipping send.')
             raise gen.Return()
 
         # Finally, send the message and check our return value
--- a/kingpin/actors/slack.py
+++ b/kingpin/actors/slack.py
@@ -1,0 +1,141 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2014 Nextdoor.com, Inc
+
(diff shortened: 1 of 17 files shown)


kingpin/actors/support/api.py
```diff
@@ -0,0 +1,338 @@
...
+    def __repr__(self):
+        return '%s(%s)' % (self.__class__.__name__, self)
+
+    def __str__(self):
+        return str(self._path)
+
+    def __cmp__(self, other):
+        return cmp(self._path, other._path)
+
+    def _replace_path_tokens(self, path, kwargs):
```
siminm: Using `kwargs` when not receiving `**kwargs` is confusing. Overall this function doesn't do anything except convert LookupError into a TypeError. It's not leveraging "self" so it's not class specific. I recommend deleting it altogether.

diranged:
@siminm, please review this now... its at a point where it could me merged. Its missing a few things that I'll add before we actually merge it.

* Retry decorator around the .fetch() calls
* Exception handling decorator that can handle different exceptions (500/501/502/401/402/403, etc) with unique behaviors. **This could be done separately, as its not required for the Slack API**

The one big question I have is about the location of the `kingpin.actors.support.api` package itself. Should it move into something like `kingpin.actors.utils.api`? I didn't want to muck too much around with the `utils.py` package right away, but it seems like it might be a better fit in there somehow.

kingpin/actors/slack.py
```diff
@@ -0,0 +1,141 @@
...
+        self.log.info('Sending message "%s" to Slack channel "%s"' %
+                      (self.option('message'), self.option('channel')))
+
+        # Check if our authentication creds are valid
+        auth_ok = yield self._slack_client.auth_test().http_post()
+        self._check_results(auth_ok)
+
+        # If we're in dry mode, bail out!
+        if self._dry:
+            self.log.info('Skipping message send (dry run..)')
```
siminm: Words "dry run" are already in our logger. I'd change this message to "API Credentials verified." and just 1 period.

kingpin/actors/support/api.py
```diff
@@ -0,0 +1,448 @@
...
+            '504': None,
+            '': exceptions.RecoverableActorFailure,
+        }
+    }
+
+    def __init__(self, client=None, headers=None):
+        self._client = client or httpclient.AsyncHTTPClient()
+        self.headers = headers
+
+    def _retry(f):
```
siminm: This doesn't belong to a RestClient class -- should be a standalone decorator.

siminm:
:+1:

PR #195, opened 2015-02-12
Bugfix: Fix parsing of unicode os.environ[] tokens at script parse time.

diranged:
I was expecting only strings, but did not account for unicode. Fixed up
the code to support strings, unicode and bools. Added tests. Also added
a script-runner test that literally parses a .json file properly and
should catch this in the future.


siminm:
:+1:

PR #206, opened 2015-02-19
Add zones to ELB automatically.

### LINKED ISSUE #205: aws.elb.RegisterInstance should add ELB to zone if its not already there
If an instance-ID is supplied thats in `us-west-1c`, but the ELB is not already part of `us-west-1c`, then the actor should add the ELB to that Zone and then retry.

siminm:
Close #205

### DIFF
--- a/kingpin/actors/aws/elb.py
+++ b/kingpin/actors/aws/elb.py
@@ -311,6 +311,19 @@ class RegisterInstance(base.AWSBaseActor):
         elb.register_instances(instances)
 
     @gen.coroutine
+    def _check_elb_zones(self, elb):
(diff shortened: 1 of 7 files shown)


kingpin/actors/aws/elb.py
```diff
@@ -311,6 +312,19 @@ def _add(self, elb, instances):
         elb.register_instances(instances)
 
     @gen.coroutine
+    def _check_elb_zones(self, elb):
```
diranged: Retry :)

diranged:
add retry... then merge.. :+1:

PR #208, opened 2015-02-23
Convert a Zone into a Region on the fly.

### LINKED ISSUE #198: AWS Actors should convert Zones into Regions
When a `zone` (eg. `us-west-1a`) is supplied to the `region` setting of Kingpin, we should simply convert it by stripping off the `a`.

siminm:
Close #198


diranged:
:+1:

PR #214, opened 2015-02-25
Add a warning regarding installation to Readme

siminm:


siminm:
Installing `requests` is not needed, but `--process-dependency-links` is still required.

Issue #213, opened 2015-02-25
Setup broken on clean install


siminm:
When installing (via `python setup.py install` or `pip install`) without any prior pip packages available, the following error is observed:

```
  File "/var/folders/9c/3zxdm6ts2n1dfcxz3p9f_4n80000gn/T/easy_install-h0_yZ2/requests-2.5.3/requests/__init__.py", line 53, in <module>
  File "/var/folders/9c/3zxdm6ts2n1dfcxz3p9f_4n80000gn/T/easy_install-3KS8rN/python-rightscale-0.1.3/temp/easy_install-HfTxWV/requests-2.5.3/requests/packages/__init__.py", line 49, in load_module
AttributeError: 'NoneType' object has no attribute 'modules'
```

The preceding stack-trace changes depending on our setup.py file -- commenting out a line that causes the error simply changes the stack-trace to fail on a line above.

Current workaround is to install `requests` library ahead of time.

PR #215, opened 2015-02-25
Fix pip install by updating python-rightscale dependency.

diranged:
Resolves issue #213.


siminm:
:+1: Quick! Merge it before it breaks

Issue #173, opened 2015-01-12
Define information passing between actors


mikhail:
Starting this thread to keep track of discussions on information passing between actors.

Here's a proposed user-facing syntax -- not discussing the back-end implementation of this yet.

``` JSON
{ "actor": "group.Sync",
    "desc": "Update Staging ELB SSL Certs.",
    "options": {
        "acts": [
            { "actor": "aws.elb.FindExpiringCert",
                "desc": "Get ELB with expiring Cert",
                "options": {
                    "region": "us-west-2",
                    "expiration_days": "30"
                },
                "output_name": "ExpiringELB"
            },
            { "actor": "aws.elb.UseCert",
                "desc": "Apply New Cert",
                "depends_on": "ExpiringELB",
                "condition": "{ExpiringELB.found}",
                "options": {
                    "name": "{ExpiringELB.name}",
                    "port": "{ExpiringELB.port}",
                    "region": "{ExpiringELB.region}",
                    "cert_name": "%NEW_CERT_NAME%"
                }
            }
        ]
    }
}
```

diranged:
We'll talk more about the syntax here.. but I don't want to make the syntax too complex. That said, a recent commit (#143) has set a framework up for passing _context_ into actors. That could be used to pass around a _global runtime context_ object that is used for sharing information between actors. In the coming weeks lets definitely nail down a design.

siminm:
How about this syntax...

Same concept as "contexts" for group actors, except it'll be "contexts-from" which defines another actor.

```json
{ "actor": "group.Sync",
  "contexts-from": { "actor": "aws.GetMetaData", "options": { "key": "instance-id" } },
  "acts": [
    { "actor": "hipchat.Message", "options": {"message": "Shutting down {instance-id}"}},
    { "actor": "system.ShutDown", }
  ]
}
```

PR #217, opened 2015-02-26
Visualize

diranged:
(turning into a pull request so i can comment on it)


kingpin/actors/base.py
```diff
@@ -308,6 +308,15 @@ def _fill_in_contexts(self, context={}, strict=True):
         # Finally, convert the string back into a dict and store it.
         self._options = json.loads(new_options_string)
 
+    def get_orgchart(self, parent=''):
+        return [
+            [{'v': str(id(self)),
```
diranged: I think its absolutely time for the BaseActors ` self.__repr__` to return useful and consistent data about themselves. Something like:

```(aws.elb.WaitUntilHealthy) <insert options['desc'] here>```

File: kingpin/actors/group.py [hunk 2]
```diff
@@ -62,6 +62,14 @@ def __init__(self, *args, **kwargs):
         # Pre-initialize all of our actions!
         self._actions = self._build_actions()
 
+    def get_orgchart(self, parent=''):
+        ret = super(BaseGroupActor, self).get_orgchart(parent=parent)
```
diranged: What exactly is happening here? Is this a hack so that you can get a single "root orgchart object"?

kingpin/actors/base.py
```diff
@@ -308,6 +308,15 @@ def _fill_in_contexts(self, context={}, strict=True):
         # Finally, convert the string back into a dict and store it.
         self._options = json.loads(new_options_string)
 
+    def get_orgchart(self, parent=''):
```
diranged: It would be great in the long run if the there was some kind of a `status` object associated with every `BaseActor`. This object could include information like ...

 * The `parent`/`child` relationship
 * Execution time of the actor
 * Tracking # of API calls potentially (lets say, if the actor did was able to do this)
 * Return values/stored data from the actor (this could go hand in hand with issue #173)

visualize
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python
```
diranged: I would love for this functionality to exist as a flag to the Kingpin script... so that a simple `dry` (or not `dry`!) run of Kingpin would automatically generate the output and help the user see it. Imagine if the generated data included representations of how long each step took, allowing for the graph to be used to investigate and optimize various parts of your deployment process.

Furthermore, it could also be used to much more easily pinpoint a failure -- it could output the HTML no matter what (as long as the `--graph` flag was supplied), and the actors that failed permanently could show up in RED, actors that failed (but `warn_on_failure=true`) could show up in YELLOW and all other actors could show up in blue or green.

That would be super useful!

diranged:
This is a seriously awesome little POC... :)

kingpin/actors/base.py
siminm: I think an actor could have `self._exec_data` dictionary with arbitrary data stored in there. Each actor would choose what to store, with some values being expected such as `_exec_data['runtime']`, success status, # api calls, or # of retry fails etc.etc.

kingpin/actors/base.py
siminm: Maybe for the `'f'` part. `'v'` is used as a unique ID for child-parent relationship. I don't want plain text to identify that because if a description of two actors is the same it would botch the diagram

kingpin/actors/group.py
siminm: This is unique to "Org Chart" behavior. The data supplied to Org Chart is not nested -- it's a single array of every single node. So this particular line defines the data for the group actor itself, such that all the child actors have a parent to point to. So not root -- any group actor.

visualize
siminm: Easily doable -- the meat of this entire script is `chart_data = top_actor.get_orgchart()` followed by injecting that data into an HTML template.

swaggy:
The best thing about this is how simple the code is. Fantastic.

mikhail:
@diranged this is ready for review

mikhail:
Can I get some feedback on this? It's blocking me on some further work

mikhail:
Since this is harmless and backwards compatible, I'm merging this now and will address any comments as they come.

Issue #222, opened 2015-03-03
New Actor: rightscale.server_array.RollingRelaunch


diranged:
It would be pretty useful to have a Kingpin actor that could walk through a SeverArray and execute a rolling-relaunch of all the hosts in that array. I could imagine it looking like this:

```json5
{ "desc": "Replace all of our WWW Hosts, two at a time",
  "actor": "rightscale.server_array.RollingRelaunch",
  "options": {
    "array": "prod-www-uswest1",
    "concurrent": 2
  }
}
```

siminm:
I'd name it just  `Relaunch`, and make `concurrent` optional. Other than that lgtm.

diranged:
Since this can actually be done purely in Kingpin JSON scripts now (see #228), closing it.

PR #228, opened 2015-03-07
Allow setting of the 'strictness' of the rightscale.server_array.Clone actor

diranged:
(Solves issue #222 without writing a whole new actor)

The only reason that a "relaunch" of an existing ServerArray couldn't be
automated with existing actors was that the
rightscale.server_array.Clone actor was extremely strict about whether
the source/destination arrays already existed or not. By allowing this
to be configurable, you can write a script that acts on not-yet-existing
arrays.

Included in this change is an example of a "relaunch" Kingpin script.


siminm:
Very nice!

Issue #116, opened 2014-10-24
Add a 'timeout' to kingpin.actors.base.BaseActor


diranged:
There should be a global way of configuring a timeout setting for every Actor that executes. If the actor takes longer than `timeout`, its *somehow* gracefully terminated and fails out. This is hard because actors could be firing off threads, which are hard to control. We need to decide if the idea is to *kill* the actor as its running, or simply *skip it* gracefully (if `warn_on_fail=True`) or *fail properly* (if `warn_on_fail=False`).

diranged:
I've begun a bit of this work. Tornado 4 support is here: https://github.com/Nextdoor/kingpin/tree/tornado_4

PR #230, opened 2015-03-09
Add Execution timeout setting to BaseActor

diranged:
Handles issue #116


siminm:
Aside from `timeout` not really being a class method, it lgtm. Any reason not to move it to `utils` library where we have everything else?

siminm:
:+1:

PR #233, opened 2015-03-11
Accept 'string counts' for rightscale.server_array.Launch.

diranged:
This allows '10' to be passed in (or more likely, using contexts,
something like '{COUNT}') and then auto converts it to an int.

### DIFF
--- a/kingpin/actors/rightscale/server_array.py
+++ b/kingpin/actors/rightscale/server_array.py
@@ -496,7 +496,7 @@ class Launch(ServerArrayBaseActor):
     all_options = {
         'array': (str, REQUIRED, 'ServerArray name to launch'),
         'count': (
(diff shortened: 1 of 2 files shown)


kingpin/actors/rightscale/server_array.py
```diff
@@ -539,7 +544,7 @@ def _wait_until_healthy(self, array, sleep=60):
         # Get the current min_count setting from the ServerArray object, or get
         # the min_count from the count number supplied to the actor (if it
         # was).
-        min_count = self._options.get('count', False)
+        min_count = int(self._options.get('count', False))
```
siminm: Shouldn't this be min_count = count_specified ?

siminm:
:+1:

PR #234, opened 2015-03-12
Authentication testing/logging improvements to actors.support.api Package

diranged:
General improvements to the testing, and logging of this package. Hides passwords/tokens in logging and reduces unnecessary debug logging.

### DIFF
--- a/kingpin/actors/support/api.py
+++ b/kingpin/actors/support/api.py
@@ -54,11 +54,26 @@ def _retry(f):
 
     def wrapper(self, *args, **kwargs):
         i = 1
+
+        # Check if the calling object has a _private_kwargs attribute.
+        private_kwargs = []
+        if hasattr(self, '_private_kwargs'):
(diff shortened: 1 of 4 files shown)


kingpin/actors/support/api.py
```diff
@@ -54,11 +54,26 @@ def _retry(f):
 
     def wrapper(self, *args, **kwargs):
         i = 1
+
+        # Check if the calling object has a _private_kwargs attribute.
+        private_kwargs = []
+        if hasattr(self, '_private_kwargs'):
+            private_kwargs = self._private_kwargs
```
siminm: Can use getattr() here with `[]` for default

siminm:
1 comment for `api.py` but other than that lgtm. Want to change the Pingdom actor to use the new class in this PR?

kingpin/actors/support/api.py
diranged: nice! Done!

Issue #232, opened 2015-03-10
rightscale.server_array.Execute does not handle 403 correctly


diranged:
When a 403 is returned, we are not handling it correctly and are instead printing out a stack trace. We should handle it and raise an InvalidCredentials exception.

```
12:12:46   DEBUG    37640 [kingpin.utils                           ] [wrapper             ] Exception caught in <function make_generic_request at 0x106458758>((<kingpin.actors.rightscale.api.RightScale object at 0x106457ad0>, u'/api/clouds/6/instances/DBMF6ISEHMJLE/run_executable'), {'post': {u'inputs[COMMAND]': u'text:uptime', u'inputs[USERNAME]': u'text:root', 'right_script_href': u'/api/right_scripts/527606003'}}): 403 Client Error: Forbidden
12:12:46   DEBUG    37640 [kingpin.utils                           ] [wrapper             ] Traceback (most recent call last):
  File "/Users/diranged/git/Nextdoor/dev-tools/kingpin/.venv/lib/python2.7/site-packages/kingpin/utils.py", line 179, in wrapper
    return func(*args, **kwargs)
  File "/Users/diranged/git/Nextdoor/dev-tools/kingpin/.venv/lib/python2.7/site-packages/kingpin/actors/rightscale/api.py", line 682, in make_generic_request
    response = self._client.client.post(url, data=post)
  File "/Users/diranged/git/Nextdoor/dev-tools/kingpin/.venv/lib/python2.7/site-packages/rightscale/httpclient.py", line 90, in request
    r.raise_for_status()
  File "/Users/diranged/git/Nextdoor/dev-tools/kingpin/.venv/lib/python2.7/site-packages/requests/models.py", line 831, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
HTTPError: 403 Client Error: Forbidden

12:12:46   CRITICAL 37640 [kingpin.actors.rightscale.server_array.Execute] [_execute_array      ] [Puppet Dry Run (env=staging)] Script execution error: Script Execution Error: 403 Client Error: Forbidden
12:12:46   ERROR    37640 [tornado.application                     ] [__del__             ] Future exception was never retrieved: Traceback (most recent call last):
  File "/Users/diranged/git/Nextdoor/dev-tools/kingpin/.venv/lib/python2.7/site-packages/tornado/gen.py", line 606, in callback
    result_list = [i.result() for i in children]
[...]

PR #236, opened 2015-03-12
Force Tornado < 4.1 due to a behavioral change in Exception logging.

diranged:
Something in the way our code is calling multiple tasks and yielding on
them is causing Tornado 4.1 to log out all the exceptions again,
regardless of whether we're doing it or not. Until this is resolved,
pinning our Tornado to < 4.1.

I have opened an issue with the Tornado team asking for guidance:
https://github.com/tornadoweb/tornado/issues/1378

This handles #232 -- where really we're handling the code fine, but there is extra logging being generated by Tornado that we were not expecting.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,13 @@
 # General App Requirements
 
-# 4.0+ is required for the @gen.with_timeout decorator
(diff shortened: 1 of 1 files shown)


siminm:
:+1:

Issue #229, opened 2015-03-07
rightscale.server_array.Execute actor should handle 422 better


diranged:
When an unexpected 422 error occurs, the Kingpin output should be more clear about exactly which host(s) failed to execute a script: Here is an example output that isn't very helpful:

```
00:37:01   CRITICAL  [Remove old tools from ELB] Script execution error: Script Execution Error: 422 Client Error: Unprocessable Entity
00:37:01   CRITICAL  [Remove old tools from ELB] Invalid parameters supplied to execute script.
00:37:01   CRITICAL  [Remove old arrays from the ELBs.] Exceptions raised by 1 of 6 actors in "Remove old arrays from the ELBs.".
00:37:01   ERROR     [STAGE 2: Promote the new deployment to be active.] Not executing any following actions because "Remove old arrays from the ELBs." failed
00:37:01   CRITICAL  [STAGE 2: Promote the new deployment to be active.] Exceptions raised by 1 of 6 actors in "Remove old arrays from the ELBs.".
00:37:01   ERROR     [Deploy to production: Django - 0000-abcdefg] Not executing any following actions because "STAGE 2: Promote the new deployment to be active." failed
00:37:01   CRITICAL  [Deploy to production: Django - 0000-abcdefg] Exceptions raised by 1 of 6 actors in "Remove old arrays from the ELBs.".
00:37:01   CRITICAL  [Kingpin] Exceptions raised by 1 of 6 actors in "Remove old arrays from the ELBs.".
00:37:01   ERROR     Kingpin encountered mistakes during the play.
00:37:01   ERROR     Exceptions raised by 1 of 6 actors in "Remove old arrays from the ELBs.".
```

* We should be able to explicitly tell you which *instance* failed.
* If the error happens immediately (on the 'queuing' of the job), we should be able to retry. If the error happens after the job is queued, then we have to simply exit... but with more output.

diranged:
Following up here. After digging, its clear that we just need to be able to output exactly which hosts failed. The problem is still an error in RightScales end, but by giving the user better output we can make it easier to recover.

diranged:
Pull #237 handles this.

PR #237, opened 2015-03-13
Support Tornado 4.1+ by ensuring all exceptions are caught and handled.

diranged:
Tornado 4.1 was much louder about exceptions that it did not believe had been handled in asynchronous Future objects. This patch cleans up our use of these Futures while also creating much more clear output to the user about which instances failed to execute a script.

This also handles issue #229, as well as #236.


siminm:
:+1:

PR #238, opened 2015-03-13
Improve the timeout system...

diranged:
A few changes here to the timeouts before we release em.

  * `group.Sync`/`group.Async` actors default to *no* timeout.
  * Added `DEFAULT_TIMEOUT` as an environment variable that Kingpin will pay attention to
  * Resolved a race condition that caused extraneous logging of ActorTimedOut exceptions.


diranged:
CC: @chuckkarish, @siminm

siminm:
:+1:

PR #239, opened 2015-03-13
Add the @sync_retry decorator to the make_generic_request() call.

diranged:
This is just a safety check. If it fails, we should at least try a few
times.


siminm:
:+1:

PR #243, opened 2015-03-21
Retry on HTTP 599s in support.api

diranged:


siminm:
:+1: 

I also think we can change the convention for "default" error to be `'*'` instead of an emptystring

PR #258, opened 2015-04-10
Set the default timeout for the misc.Macro actor to 0 (unlimited).

diranged:
Just like the Group actors, the Macro actor does nothing but call other
actors. Its timeout should be 0 by default.


diranged:
CC: @chuckkarish, @bemsha, @siminm

nextdoor-engineer:
:+1:

PR #259, opened 2015-04-19
Allow Slack message formatting

siminm:


siminm:
I think it's the opposite. I've read that documentation and tested it by hand. Here's an excerpt, 

> If you want Slack to treat your message as completely unformatted, pass parse=full. This will ignore any markup formatting you added to your message.

diranged:
After discussing, it seems that the documentation in Slacks API is wrong (or misleading). This change is fine. :+1:

PR #263, opened 2015-04-24
 Clean up all AWS Actors: Use retrying.retry with common backoff settings...

diranged:
....

The AWS API is pretty sensitive when it comes to hammering it with
requests. We had all kinds of different retry logic in our code here as
we worked to deal with that.

This change wipes out all of that logic and puts it into a single
location -- the AWSBaseActor.thread() method. This provides us a single
common location to handle retrying. With this, we no longer use the
kingpin.actors.support.api._retry() method (which was sort of abused
here anyways).

The retrying logic is described in the kingpin.actors.aws.settings
module ... but has significant jitter, randomness, and exponential
backoffs applied.

### DIFF
--- a/kingpin/actors/aws/base.py
+++ b/kingpin/actors/aws/base.py
@@ -17,20 +17,22 @@
 import logging
 import re
 
-from boto import utils
-from boto.exception import BotoServerError
+from boto import utils as boto_utils
+from boto import exception as boto_exception
 from tornado import concurrent
 from tornado import gen
 from tornado import ioloop
+from retrying import retry
+import boto.cloudformation
 import boto.ec2
 import boto.ec2.elb
 import boto.iam
+import boto.sqs
 
+from kingpin import utils
 from kingpin.actors import base
 from kingpin.actors import exceptions
 from kingpin.actors.aws import settings as aws_settings
-
-from kingpin.actors.support import api as support
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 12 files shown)


kingpin/actors/aws/base.py
```diff
@@ -120,7 +112,14 @@ def __init__(self, *args, **kwargs):
             aws_access_key_id=aws_settings.AWS_ACCESS_KEY_ID,
             aws_secret_access_key=aws_settings.AWS_SECRET_ACCESS_KEY)
 
+        self.cf_conn = boto.cloudformation.connect_to_region(
+            region,
+            aws_access_key_id=aws_settings.AWS_ACCESS_KEY_ID,
+            aws_secret_access_key=aws_settings.AWS_SECRET_ACCESS_KEY)
+
     @concurrent.run_on_executor
+    @retry(**aws_settings.RETRYING_SETTINGS)
```
siminm: What do you think about `@retry(**self._retrying_settings)` ?

diranged:
CC: @bemsha: This should reduce or eliminate the API throttling issues we were seeing. Specifically, the new AWS retry settings are in the `kingpin/actors/aws/settings.py` file now for you to read.

siminm:
:+1: yay

bemsha:
I like. Cleanliness wins.

      Chuck Karish             chuck@nextdoor.com               650 283 7864

PR #264, opened 2015-04-26
Use self.thread() in elb.WaitUntilHealthy._is_healthy() method.

diranged:
We were making an API call here with no retry decorator at all. Oops!
Now we're using the self.thread() method to handle the API call and deal
with all the retry logic.


nextdoor-engineer:
:+1:

PR #266, opened 2015-04-29
Handle RightScales concept of 'array values'.

diranged:
Per
http://reference.rightscale.com/api1.5/resources/ResourceServerArrays.html#create_parameters,
and
http://support.rightscale.com/12-Guides/RightScale_API_1.5/1_Overview/Parameters/index.html:

RightScale has a concept of input 'arrays' that are passed in as funny
looking key=value pairs. For example, passing in a schedule for
AutoScaling an array in RightScale looks like this:

  server_array[elasticity_params][schedule][][min_count]=1
  server_array[elasticity_params][schedule][][day]=Sunday
  server_array[elasticity_params][schedule][][max_count]=2
  server_array[elasticity_params][schedule][][time]=0700

  server_array[elasticity_params][schedule][][max_count]=2
  server_array[elasticity_params][schedule][][day]=Monday
  server_array[elasticity_params][schedule][][min_count]=1
  server_array[elasticity_params][schedule][][time]=0700

It should be noted that order only partially matters here. As long as
each group of 'day', 'max_count', 'min_count' and 'time' values are put
together, the specific order they are in does not matter. Note the
difference between the first 4 items and the second 4 items in the list
above.

This is extremely un-intuitive, but appears to be the way RightScale
works.

### DIFF
--- a/kingpin/actors/rightscale/base.py
+++ b/kingpin/actors/rightscale/base.py
@@ -84,29 +84,41 @@ class RightScaleBaseActor(base.BaseActor):
 
         We return:
 
-            {'server_array[name]': 'unittest-name',
-             'server_array[bounds][min_count]': 3}
+            [ ('server_array[name]', 'unittest-name'),
+              ('server_array[bounds][min_count]', '3) ]
+
+        For more examples, see our unit tests.
 
         Args:
             prefix: The key-prefix to use (ie, 'server_array')
             params: The dictionary to squash
 
         Returns:
-            A single-level dictionary of key/value pairs.
+            A list of tuples of key/value pairs.
         """
         if not type(params) == dict:
             raise exceptions.InvalidOptions(
                 'Parameters passed in must be in the form of a dict.')
 
         # Nested loop that compresses a multi level dictinary into a flat
-        # dict of key/value pairs.
+        # array of key=value strings.
         def flatten(d, parent_key=prefix, sep='_'):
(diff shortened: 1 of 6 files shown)


kingpin/actors/rightscale/base.py
```diff
@@ -84,29 +84,39 @@ def _generate_rightscale_params(self, prefix, params):
...
-                    items.extend(flatten(v, new_key).items())
-                else:
-                    items.append((new_key, v))
-            return dict(items)
+
+            if isinstance(d, collections.MutableMapping):
+                # If a dict is passed in, break it into its items and
+                # then iterate over them.
+                for k, v in d.items():
+                    new_key = parent_key + '[' + k + ']' if parent_key else k
```
siminm: This isn't intuitive -- either refactor this, or add an example comment

kingpin/actors/rightscale/base.py
```diff
@@ -84,29 +84,39 @@ def _generate_rightscale_params(self, prefix, params):
...
+                    items.extend(flatten(v, new_key))
+            elif isinstance(d, list):
+                # If an array was passed in, then iterate over the array
+                new_key = parent_key + '[]' if parent_key else k
+                for item in d:
+                    items.extend(flatten(item, new_key))
```
siminm: Same here

siminm:
:+1:

PR #267, opened 2015-04-30
Allow contexts to be stored in a separate file.

siminm:
This change allows putting a list of ELBs / Arrays that we operate on into a separate file -- so that various script files can have a single source of information.

cc: @diranged @bemsha


diranged:
High level, this is a reasonable idea. Two implementation details..

* I would rather see the `context` parameter be smart enough to figure out if someone supplied a string (thus, a reference to a file) and automatically read the file in.
* Can't we re-use our existing JSON-reader code so that we're not calling `open()`?

PR #269, opened 2015-05-04
Disable the cross-actor RightScale API Object sharing.

diranged:
The python-rightscale library is supposed to
(https://github.com/brantai/python-rightscale/commit/2a8c0f572c4b13de03cbf67a1253c135b45ff2ca)
honor the RightScale OAUTH2 token expiration date and generate a new
token when it expires. This code worked when I wrote it and individually
tested it inside the confines of that library itself.

However, we are still seeing cases where extremely long running
processes in Kingpin will hit a 403 error from RightSCale roughly 2
hours after the initial OAUTH token was created. Given the complexity of
firing off all of our API calls into little mini threads everywhere, I'm
concerned that there is some kind of non-thread-safe code issue here.

To simplify things, I'm removing the cross-actor RightScale API object
sharing. Each actor will instantiate its own RightScale API object and
that object will get its own fresh OAUTH token. At the bare minimum this
means that the 2-hour timeout applies to each individual actor, rather
than the entire execution of the Kingpin script.

The long term fix here is likely to write a fresh API client for
RightScale based on our kingpin.actors.support.api module.

### DIFF
--- a/kingpin/actors/rightscale/base.py
+++ b/kingpin/actors/rightscale/base.py
@@ -35,8 +35,6 @@ class RightScaleBaseActor(base.BaseActor):
 
     """Abstract class for creating RightScale cloud actors."""
 
-    CLIENTS = {}
-
     def __init__(self, *args, **kwargs):
         """Initializes the Actor."""
         super(RightScaleBaseActor, self).__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


siminm:
:+1: 1:

PR #270, opened 2015-06-12
RightScale Alert Spec Actors

diranged:
This is still a WIP ... Opening the PR to easily compare the diff and share with the team.


diranged:
@siminm, @pjanakiraman, @chuckkarish: New RightScale AlertSpec creation actors... yay!

diranged:
@siminm: Please review..

kingpin/actors/rightscale/alerts.py
```diff
@@ -0,0 +1,358 @@
...
+            name: RightScale AlertSpec Name
+            subject_href: The HREF of the subject this AlertSpec is assigned
+              to.
+
+        Return:
+            [<rightcale.Resource objects>]
+        """
+        log.debug('Searching for AlertSpec matching: %s' % name)
+        found_spec = yield self._client.find_by_name_and_keys(
+            self._client._client.alert_specs, name, exact=False,
```
siminm: Why exact `False`? If there's a legitimate use case for this then please document it above

kingpin/actors/rightscale/alerts.py
```diff
@@ -0,0 +1,358 @@
...
+    the Alert Spec that it would have created.
+
+    Example *dry* output::
+
+        TODO: Fill this in
+    """
+
+    all_options = {
+        'array': (str, REQUIRED, 'Name of the ServerArray act on.'),
+        'strict_array': (bool, False, 'Strict ServerArray validation.'),
```
siminm: This description doesn't match the Sphinx description above.

Also what's the behavior if the array is not found?

kingpin/actors/rightscale/alerts.py
diranged: This is intentional -- but I added docs.

kingpin/actors/rightscale/alerts.py
diranged: Docs updated. It raises a standard ServerArrayNotFound exception.

kingpin/actors/rightscale/alerts.py
```diff
@@ -0,0 +1,364 @@
...
+            log.debug('AlertSpec matching "%s" could not be found.' % name)
+            return
+
+        log.debug('Got AlertSpec: %s' % found_spec)
+        raise gen.Return(found_spec)
+
+
+class Create(AlertsBaseActor):
+
+    """Create a RightScale Alert Spec
```
siminm: Add a link to RightScale API describing the expected parameters. Seems that our option names are identical to the ones required by the API

kingpin/actors/rightscale/alerts.py
```diff
@@ -0,0 +1,364 @@
...
+        # if their option value came in as None.
+        _optional_params = [
+            'description', 'escalation_name', 'vote_tag', 'vote_type'
+        ]
+        for optional in _optional_params:
+            if self.option(optional):
+                params[optional] = self.option(optional)
+
+        params = self._generate_rightscale_params('alert_spec', params)
+        self.log.debug('Generated params: %s' % params)
```
siminm: Doesn't _generate_rightsacle_params have a debug log already?

kingpin/actors/rightscale/alerts.py
```diff
@@ -0,0 +1,364 @@
...
+
+        if self._dry:
+            # In dry run mode, just log out what we would have done.
+            self.log.info('Would have created the alert spec \"%s\" on %s' %
+                          (self.option('name'), array.soul['name']))
```
siminm: I recommend exiting here `raise gen.Return()` and not using an `else` at all

kingpin/actors/rightscale/alerts.py
```diff
@@ -0,0 +1,364 @@
...
+        else:
+            # We're really doin this. If we get a known exception back, handle
+            # it. Otherwise, raise it.
+            try:
+                yield self._client.create_resource(
```
siminm: This should be split into its own function, wrapped in `@retry` -- so that HTTPError is handled automatically

kingpin/actors/rightscale/alerts.py
diranged: Done

kingpin/actors/rightscale/alerts.py
diranged: Nope

kingpin/actors/rightscale/alerts.py
diranged: Done

siminm:
:+1: sexy time!

kingpin/actors/rightscale/alerts.py
diranged: Per our discussion -- not making changes here. The `self._client.create_resource()` method has its own `@retry` decorator already.

PR #273, opened 2015-06-20
Allow Async actor to limit concurrency

siminm:
cc: @diranged

### DIFF
--- a/kingpin/actors/group.py
+++ b/kingpin/actors/group.py
@@ -273,6 +273,11 @@ class Async(BaseGroupActor):
 
     **Options**
 
+    :concurrency:
+      Max number of concurrent executions. This will fire off N executions
+      in parallel, and continue with the remained as soon as the first
+      execution is done. This is faster than creating N Sync executions.
+
     :acts:
       An array of individual Actor definitions.
 
@@ -328,6 +333,12 @@ class Async(BaseGroupActor):
     finish before the failure is returned.
     """
 
+    all_options = {
+        'concurrency': (int, 0, "Max number of concurrent executions."),
+        'contexts': (list, [], "List of contextual hashes."),
+        'acts': (list, REQUIRED, "Array of actor definitions.")
+    }
+
     @gen.coroutine
     def _run_actions(self):
         """Asynchronously executes all of the Actor.execute() methods.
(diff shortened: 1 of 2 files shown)


kingpin/actors/group.py
```diff
@@ -328,6 +333,24 @@ class Async(BaseGroupActor):
...
+    def worker(self, act_gen, worker_id):
+        """Syncronous actor-executor that fetches from a shared generator.
+
+        Concurrency specifies the total number of workers. Each worker
+        continuously grabs the next task (actor) and executes it. Since the
+        acts come from the same pool (generator) no actions are executed twice.
+        """
+        for act in act_gen:
+            self.log.debug('Worker[%s]: %s' % (worker_id, act._desc))
+            yield act.execute()
```
siminm: Todo: figure out a fix for exception handling, such that the end-result is same as `yield [task, task, task]`

siminm:
Reworked the logic to be simpler, and consistent with the rest of Tornado yield/raise error handling. No need to make this a separate module

kingpin/actors/group.py
```diff
@@ -338,14 +349,18 @@ def _run_actions(self):
         failed (False).
         """
 
-        # This is an interesting tornado-ism. Here we generate and fire off
```
diranged: I wouldn't remove this comment -- its still very useful.

kingpin/actors/group.py
```diff
@@ -338,14 +349,18 @@ def _run_actions(self):
...
         tasks = []
         for act in self._actions:
             tasks.append(act.execute())
 
+            if self.option('concurrency') > 0:
+                running_tasks = len([t for t in tasks if t.running()])
+
+                while running_tasks == self.option('concurrency'):
+                    self.log.debug('Concurrecnty saturated. Waiting.')
+                    yield gen.sleep(0)
```
diranged: I assume you want to sleep longer, right?

kingpin/actors/group.py
```diff
@@ -338,14 +349,18 @@ def _run_actions(self):
...
-        # references to those tasks. However, we don't yield (wait) on them to
-        # finish.
         tasks = []
         for act in self._actions:
             tasks.append(act.execute())
 
+            if self.option('concurrency') > 0:
+                running_tasks = len([t for t in tasks if t.running()])
+
+                while running_tasks == self.option('concurrency'):
```
diranged: Just in case ... use `>=`?

kingpin/actors/group.py
siminm: done

kingpin/actors/group.py
siminm: done

kingpin/actors/group.py
siminm: Ideally I'd have some `noop()` function. I don't actually need to sleep, I just need to allow the tornado stack to progress due to the internal ioloop. Basically making this equivalent to "do *something* then check again"

siminm:
Addressed the comments and our IRL discussion. Ready for review again

diranged:
:+1:

diranged:
Would you consider adding this to the `rightscale.Execute` actor?

siminm:
yay http://kingpin.readthedocs.org/en/latest/kingpin.actors.group.html?highlight=concurrency

PR #276, opened 2015-06-25
New Actor: rightscale.server_array.UpdateNextInstance

diranged:
Sometimes you want to update the 'Next Instance' settings of a
RightScale Array -- these include things like the Server Template, IP
Association settings, AMI Image HREFs and more.

This actor searches for your ServerArray, grabs its 'Next Instance'
object, and then updates that object with the parameters you've passed
in.

Additionally, a custom helper has been added so that if you use
'default' as your 'image_href', we automatically discover the default
AMI image for your MCI and associate that image_href for you.

CC: @siminm

### DIFF
--- a/kingpin/actors/rightscale/api.py
+++ b/kingpin/actors/rightscale/api.py
@@ -123,6 +123,7 @@ class RightScale(object):
         return int(path.split(resource.self.path)[-1])
 
     @concurrent.run_on_executor
+    @sync_retry(**settings.RETRYING_SETTINGS)
     @utils.exception_logger
     def find_server_arrays(self, name, exact=True):
         """Search for a list of ServerArray by name and return the resources.
@@ -152,6 +153,20 @@ class RightScale(object):
         log.debug('Got ServerArray(s): %s' % ', '.join(names))
 
         return found_arrays
+
+    @concurrent.run_on_executor
+    @sync_retry(**settings.RETRYING_SETTINGS)
+    @utils.exception_logger
+    def show(self, resource):
+        """Async wrapping of <resource>.show() with retry wrapper.
+
+        Args:
+            resource: rightscale.Resource object
+
+        Returns:
+            <rightscale.Resource object>.show()
+        """
+        return resource.show()
 
     @concurrent.run_on_executor
     @utils.exception_logger
@@ -186,6 +201,7 @@ class RightScale(object):
         return recipe
 
     @concurrent.run_on_executor
+    @sync_retry(**settings.RETRYING_SETTINGS)
     @utils.exception_logger
     def find_right_script(self, name):
         """Search for a RightScript by-name and return the resource.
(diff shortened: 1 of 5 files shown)


kingpin/actors/rightscale/api.py
```diff
@@ -327,11 +345,11 @@ def update_server_array(self, array, params):
...
 
-        log.debug('Patching ServerArray (%s) with new params: %s' %
-                  (array.soul['name'], params))
-        array.self.update(params=params)
-        updated_array = array.self.show()
-        return updated_array
+        log.debug('Patching %s with new params: %s' %
+                  (resource.soul['name'], params))
+        resource.self.update(params=params)
+        updated_resource = resource.self.show()
```
siminm: I think you mean `updated_resource = self.show(resource.self)`

kingpin/actors/rightscale/server_array.py
```diff
@@ -421,6 +421,187 @@ def _execute(self):
...
+            self.option('array'), exact=self.option('exact'))
+
+        # In dry run, just comment that we would have made the change.
+        if self._dry:
+            self.log.debug('Not making any changes.')
+            self.log.info('Params would be: %s' % self.option('params'))
+            raise gen.Return()
+
+        # Do the real work
+        yield self._apply(self._update_params, arrays)
```
siminm: What's `_apply`? Also I think we've discussed in person that _execute should do much of the heavy lifting rather than calling another function that does everything

kingpin/actors/rightscale/api.py
diranged: Not really -- this is a synchronous method wrapped in `@concurrent.run_on_executor`, just like `self.show()` is. The only change I really made here was to make this not array-specific, and let it be re-used for other non-array resources that take the `params` arg.

kingpin/actors/rightscale/server_array.py
diranged: `self._apply()` is a helper method in the `ServerArrayBaseActor` that allows us to quickly execute a method across a group of arrays rather than a single one. Its been there for a little while.

As for the complication question.. I agree entirely. However, I did not want to write this particular actor differently than the rest of the RightScale actors. I chose here to keep a common style within the `rightscale.server_array.*` actors.

diranged:
@siminm, I didn't address your comments with code-changes (bt I did comment on them). I also made a quick change to the docstring to deal with a line length issue.

kingpin/actors/rightscale/api.py
siminm: Then what's the point of `def show(...)` above?

siminm: Scratch that, I keep getting confused between the api and the actor. I see it's used in the actor

siminm:
:+1:

PR #277, opened 2015-07-03
Packagecloud Actor

cmclaughlin:
@siminm @diranged  Review please.  No tests so far, so just let me know if this looks good so far and if so, I'll add tests.

### DIFF
--- a/kingpin/actors/base.py
+++ b/kingpin/actors/base.py
@@ -511,7 +511,8 @@ class HTTPBaseActor(BaseActor):
     # garbled data (ie, maybe a 500 errror or something else thats not in
     # JSON format, we should back off and try again.
     @gen.coroutine
-    def _fetch(self, url, post=None, auth_username=None, auth_password=None):
+    def _fetch(self, url, post=None, auth_username=None, auth_password=None,
+               method='GET'):
         """Executes a web request asynchronously and yields the body.
 
         Args:
@@ -519,16 +520,22 @@ class HTTPBaseActor(BaseActor):
             post: (Str) POST body data to submit (if any)
             auth_username: (str) HTTP auth username
             auth_password: (str) HTTP auth password
+            method: HTTP method
         """
 
         # Generate the full request URL and log out what we're doing...
         self.log.debug('Making HTTP request to %s with data: %s' % (url, post))
 
+        if not method and post:
+            method = 'POST'
+        elif not method and not post:
+            method = 'GET'
+
         # Create the http_request object
         http_client = self._get_http_client()
         http_request = httpclient.HTTPRequest(
             url=url,
-            method=self._get_method(post),
+            method=method,
             body=post,
             headers=self.headers,
             auth_username=auth_username,
--- a/kingpin/actors/base.py
+++ b/kingpin/actors/base.py
@@ -511,8 +511,7 @@ class HTTPBaseActor(BaseActor):
     # garbled data (ie, maybe a 500 errror or something else thats not in
     # JSON format, we should back off and try again.
     @gen.coroutine
-    def _fetch(self, url, post=None, auth_username=None, auth_password=None,
-               method='GET'):
+    def _fetch(self, url, post=None, auth_username=None, auth_password=None):
         """Executes a web request asynchronously and yields the body.
 
         Args:
@@ -520,22 +519,16 @@ class HTTPBaseActor(BaseActor):
             post: (Str) POST body data to submit (if any)
             auth_username: (str) HTTP auth username
             auth_password: (str) HTTP auth password
-            method: HTTP method
         """
 
         # Generate the full request URL and log out what we're doing...
         self.log.debug('Making HTTP request to %s with data: %s' % (url, post))
 
-        if not method and post:
-            method = 'POST'
-        elif not method and not post:
-            method = 'GET'
-
         # Create the http_request object
         http_client = self._get_http_client()
         http_request = httpclient.HTTPRequest(
             url=url,
-            method=method,
+            method=self._get_method(post),
             body=post,
             headers=self.headers,
             auth_username=auth_username,
--- a/kingpin/actors/packagecloud.py
+++ b/kingpin/actors/packagecloud.py
@@ -124,7 +124,7 @@ class PackagecloudBase(base.BaseActor):
             'name': package['name']
         } for package in all_packages if package['name'] == name]
 
-        versions.sort(key=lambda x: x.get('created_at'), reverse=True)
+        versions.sort(key=lambda x: x.get('created_at'), reverse=False)
         return versions
 
     @gen.coroutine
@@ -139,9 +139,8 @@ class PackagecloudBase(base.BaseActor):
             A list of unique package names that match the delete pattern.
             """
         pattern = re.compile(packages_to_delete)
-        packages_list_to_delete = [package['name'] for package in all_packages
-                                   if pattern.match(package['name'])]
-        packages_list_to_delete = set(packages_list_to_delete)
+        packages_list_to_delete = {package['name'] for package in all_packages
+                                   if pattern.match(package['name'])}
 
         self.log.debug('List of packages matching regex (%s): %s' %
                        (packages_to_delete, packages_list_to_delete))
@@ -166,31 +165,51 @@ class PackagecloudBase(base.BaseActor):
         all_packages = yield self._get_all_packages(repo=repo)
         packages_list_to_delete = yield self._get_packages_list_to_delete(
             packages_to_delete, all_packages)
-        deleted_packages = []
 
         # Loop through each unique package to delete
         for name in packages_list_to_delete:
             package_versions = self._get_package_versions(
                 name, all_packages)
 
-            # Delete individual packages if they meet our criteria
-            number_in_repo = len(package_versions)
-            for package in package_versions:
+            # Create a tally of the packages we delete -- usd to give the user
+            # a final helpful log statement about the work we did.
+            packages_deleted = []
 
+            # Get a total count of the number of versions of this package in
+            # the repo -- this variable will then be counted down as we loop,
+            # to prevent us from deleting more than 'number_to_keep' packages.
+            number_in_repo = len(package_versions)
+            self.log.debug('Scanning %s versions (%s)' %
+                           (name, number_in_repo))
+
+            for package in package_versions:
                 # Safety check -- if there aren't more than the number_to_keep
                 # in the repo, then don't bother continuing through the loop
                 # for this package. Break out and move to the next name in
                 # packages_list_to_delete.
                 if number_in_repo <= number_to_keep:
+                    self.log.debug(
+                        '%s has only %s package versions left, skipping'
+                        % (name, number_in_repo))
                     break
 
+                # If older_than (time in seconds) was supplied, figure out how
+                # old the package is. If the package_age (in seconds) is
+                # younger than allowed_age (in seconds), then skip to the next
+                # package version in the set.
                 if older_than:
                     package_age = (datetime.datetime.now() -
                                    package['created_at'])
                     allowed_age = datetime.timedelta(seconds=older_than)
                     if package_age <= allowed_age:
+                        self.log.debug(
+                            '%s/%s is only %s old, skipping deletion' %
+                            (package['distro_version'], package['name'], package_age))
                         continue
 
+                # Finally if we got here, then we have enough packages left in
+                # the repo, AND (optionally) this package is older than our
+                # cutoff age... so delete the package.
                 msg = '%s/%s/%s' % (
                     repo, package['distro_version'], package['filename'])
                 if self._dry:
@@ -206,10 +225,23 @@ class PackagecloudBase(base.BaseActor):
 
                 # Decrement list of packages to track how many are left
                 number_in_repo = number_in_repo - 1
-                # Keeping track of removed packages for return value
-                deleted_packages.append(package)
 
-        raise gen.Return(deleted_packages)
+                # Track that this package was deleted -- used in the parent for
+                # loop to give the user a final tally of the packages that
+                # were kept, and that were deleted.
+                packages_deleted.append(package)
+
+            # Print out the packages that were not deleted and left in the repo
+            all_files = ['%s/%s' %
+                         (package['distro_version'], package['filename'])
+                         for package in package_versions]
+            deleted_files = ['%s/%s' %
+                             (package['distro_version'], package['filename'])
+                             for package in packages_deleted]
+            files_left = list(set(all_files) - set(deleted_files))
+            self.log.debug('%s remaining packages: %s' % (name, files_left))
+
+        raise gen.Return()
 
 
 class Delete(PackagecloudBase):
@@ -275,11 +307,10 @@ class Delete(PackagecloudBase):
     @gen.coroutine
     def _execute(self):
         """Deletes all packages that match the `packages_to_delete` pattern"""
-        deleted_packages = yield self._delete(
+        yield self._delete(
             packages_to_delete=self.option('packages_to_delete'),
             number_to_keep=self.option('number_to_keep'),
             repo=self.option('repo'))
-        raise gen.Return(deleted_packages)
 
 
 class DeleteByDate(PackagecloudBase):
@@ -336,12 +367,11 @@ class DeleteByDate(PackagecloudBase):
 
     @gen.coroutine
     def _execute(self):
-        deleted_packages = yield self._delete(
+        yield self._delete(
             packages_to_delete=self.option('packages_to_delete'),
             number_to_keep=self.option('number_to_keep'),
             older_than=self.option('older_than'),
             repo=self.option('repo'))
-        raise gen.Return(deleted_packages)
 
 
 class WaitForPackage(PackagecloudBase):
--- a/kingpin/actors/packagecloud.py
+++ b/kingpin/actors/packagecloud.py
@@ -1,0 +1,416 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2014 Nextdoor.com, Inc
+
+"""
+:mod:`kingpin.actors.packagecloud`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The packagecloud actor allows you to perform maintenance operations on
+repositories hosted by packagecloud.io using their API:
+
+https://packagecloud.io/docs/api
+
+**Required Environment Variables**
+
+:PACKAGECLOUD_ACCOUNT:
+  packagecloud account name, i.e. https://packagecloud.io/PACKAGECLOUD_ACCOUNT
+
+:PACKAGECLOUD_TOKEN:
+  packagecloud API Token
+"""
+
+import datetime
+import logging
+import os
+import re
+import sys
+
+from tornado import gen
+from tornado_rest_client import api
+
+from kingpin.actors import base
+from kingpin.actors import exceptions
+from kingpin.constants import REQUIRED
+
+log = logging.getLogger(__name__)
+
+__author__ = 'Charles McLaughlin <charles@nextdoor.com>'
+
+ACCOUNT = os.getenv('PACKAGECLOUD_ACCOUNT', None)
+TOKEN = os.getenv('PACKAGECLOUD_TOKEN', None)
+
+
+class PackagecloudAPI(api.RestConsumer):
+
+    ENDPOINT = 'https://packagecloud.io/api/v1/'
+    CONFIG = {
+        'attrs': {
+            'packages': {
+                'path': ('repos/%account%/%repo%/packages.json'
+                         '?per_page={}'.format(sys.maxint)),
+                'http_methods': {'get': {}}
+            },
+            'delete': {
+                'path': 'repos/%account%/%repo%/%distro_version%/%filename%',
+                'http_methods': {'delete': {}}
+            },
+        },
+        'auth': {
+            'user': TOKEN,
+            'pass': ''
+        }
+    }
+
+
+class PackagecloudBase(base.HTTPBaseActor):
+
+    """Simple packagecloud Abstract Base Object"""
+
+    def __init__(self, *args, **kwargs):
+        """Check required environment variables."""
+        super(PackagecloudBase, self).__init__(*args, **kwargs)
+
+        if not ACCOUNT:
+            raise exceptions.InvalidCredentials(
+                'Missing the "PACKAGECLOUD_ACCOUNT" environment variable.')
+
+        if not TOKEN:
+            raise exceptions.InvalidCredentials(
+                'Missing the "PACKAGECLOUD_TOKEN" environment variable.')
+
+        rest_client = api.RestClient()
+        self._packagecloud_client = PackagecloudAPI(client=rest_client)
+
+    @gen.coroutine
+    def _get_all_packages(self, repo):
+        """Simple method for fetching a dictionary of all packages in a repo
+
+        Args:
+            repo: name of the packagecloud repo to fetch from
+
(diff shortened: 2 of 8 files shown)


kingpin/actors/base.py
```diff
@@ -511,7 +511,8 @@ def _generate_escaped_url(self, url, args):
     # garbled data (ie, maybe a 500 errror or something else thats not in
     # JSON format, we should back off and try again.
     @gen.coroutine
-    def _fetch(self, url, post=None, auth_username=None, auth_password=None):
+    def _fetch(self, url, post=None, auth_username=None, auth_password=None,
+               method='GET'):
```
cmclaughlin: I could undo this since I switched to the rest client, but I figure I might as well leave it in.

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,256 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The packagecloud actor allows you to delete packages from a packagecloud repo
+or check for existance. The actor supports dry mode by looking for packages to
+delete, but leaving them intact.
+
+**Required Environment Variables**
+
+:PACKAGECLOUD_ACCOUNT:
+  packagecloud account name, i.e. https://packagecloud.io/accountname
```
diranged: Here the example should specify that `accountname` is the bit you want to set, not the entire string. Right?

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,256 @@
...
+
+        if not TOKEN:
+            raise exceptions.InvalidCredentials(
+                'Missing the "PACKAGECLOUD_TOKEN" environment variable.')
+
+        rest_client = api.RestClient()
+        self._packagecloud_client = PackagecloudAPI(client=rest_client)
+
+    @gen.coroutine
+    def _get_all_packages(self):
```
diranged: Docstrings

kingpin/actors/base.py
siminm: :+1:

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,256 @@
...
+#
+# Copyright 2014 Nextdoor.com, Inc
+
+"""
+:mod:`kingpin.actors.packagecloud`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The packagecloud actor allows you to delete packages from a packagecloud repo
+or check for existance. The actor supports dry mode by looking for packages to
+delete, but leaving them intact.
```
siminm: I'd add a link to their API documentation

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,256 @@
...
+        'auth': {
+            'user': TOKEN,
+            'pass': ''
+        }
+    }
+
+
+class PackagecloudBase(base.HTTPBaseActor):
+
+    """Simple packagecloude Abstract Base Object"""
```
siminm: I don't think they're Czech :) Just cloud.

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,256 @@
...
+                            account=ACCOUNT,
+                            repo=self.option('repo'),
+                            distro_version=package['distro_version'],
+                            filename=package['filename']
+                        ).http_delete()
+
+
+class Delete(PackagecloudBase):
+
+    """packagecloude Delete Base Object"""
```
siminm: This is no longer a "base" object -- it's the real actor.

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,256 @@
...
+
+    """packagecloude WaitForPackage Base Object"""
+
+    all_options = {
+        'name': (
+            str, REQUIRED,
+            'Name of the package to search for'),
+        'version': (
+            str, '.*',
+            'Version of the package to search for'),
```
siminm: This comment should indicate that it's a regex

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,256 @@
...
+            log.debug('Found all packages: %s' % all_packages)
+
+            version_pattern = re.compile(self.option('version'))
+            found = False
+
+            for package in all_packages:
+                if package['name'] == self.option('name'):
+                    if version_pattern.match(package['version']):
+                        log.info('Found package %s' % package)
+                        found = True
```
siminm: Skip the nested loop and the double break:

    matched_packages = [p for p in package
        if p['name'] == self.option('name') and version_pattern.match(p['version'])]

    if len(matched_packages) > 0:
        raise gen.Return()

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,256 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2014 Nextdoor.com, Inc
+
+"""
+:mod:`kingpin.actors.packagecloud`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The packagecloud actor allows you to delete packages from a packagecloud repo
```
diranged: This is is generic high level module documentation, its not yet specific to a given actor. You're documentation here sounds like `packagecloud` is an actor itself, but its not. I would re-phrase this to talk about the _Packagecloud.io actors_, and how they are generally used and set up. From there, the individual classes can have their own usage documentation.

kingpin/actors/packagecloud.py
diranged: Also, you're example (or defaults) should be wrapped in stars to properly bring attention to them in the docs. See example here: https://github.com/Nextdoor/kingpin/blob/master/kingpin/actors/slack.py#L31

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,256 @@
...
+import datetime
+import logging
+import os
+import re
+
+from tornado import gen
+
+from kingpin.actors import base
+from kingpin.actors import exceptions
+from kingpin.actors.support import api
```
diranged: We're going to deprecate the use of this module, and instead begin using the [published-version](https://github.com/Nextdoor/tornado_rest_client/) of it (http://tornado-rest-client.readthedocs.org). Would you mind doing that here? You'll need to make changes to the `requirements.txt` file, and a few subtle changes to this module, but it will mean less work for me when I go back and make the changes to the other actors.

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,256 @@
...
+
+        if not ACCOUNT:
+            raise exceptions.InvalidCredentials(
+                'Missing the "PACKAGECLOUD_ACCOUNT" environment variable.')
+
+        if not TOKEN:
+            raise exceptions.InvalidCredentials(
+                'Missing the "PACKAGECLOUD_TOKEN" environment variable.')
+
+        rest_client = api.RestClient()
```
diranged: I see no changes to the default `RestClient` for exception handling. Have you looked at the PackageCloud failure return codes and verified that the defaults will work appropriately?

https://tornado-rest-client.readthedocs.org/en/latest/basic_client.html#exception-handling-in-http-requests

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    @gen.coroutine
+    def _execute(self):
+        """Deletes all packages that match the `packages_to_delete` pattern and
+        where created before `older_than` number of seconds
+
+        raises: gen.Return()
+        """
+        yield self._delete(by_date=True)
```
diranged: The general pattern we've been following is to make the `execute()` methods have the raw _logic_ for the actor, and then just build supporting private methods that can be re-used and well tested. So for example, you may have the following private methods:

```
def _find_package(pattern, repo):
  """Returns a list of packages (with metadata about each one) that matches *pattern*"""
```

```
def _find_by_date(pattern, repo, age=5, minimum=1):
  """Finds a list of packages that match *pattern*, and are older than *age* days old, while keeping *minimum* no matter what."""
  packages = yield self._find_packages(pattern)
  for package in packages:
    ... do stuff
  raise gen.Return(deletable_packages)
```

```
def _delete_package(name, repo, version=None):
  """Deletes a package that matches *version*, or all packages if no *version* is supplied."""
```

Then you're `execute()` method looks something like:
```
def _execute(self):
  to_delete = yield self._find_by_date(options.name, options.older_than, options.number_to)keep)
  if self._dry:
    log something..
  yield self._delete(to_delete)
```

kingpin/actors/packagecloud.py
cmclaughlin: There are no defaults for PACKAGECLOUD_ACCOUNT and PACKAGECLOUD_TOKEN -- those will be specific to every site.

kingpin/actors/packagecloud.py
cmclaughlin: Their API docs doesn't list any special return codes other than what you've listed in rest client.

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,358 @@
...
+        """Simple method for fetching a dictionary of all packages in a repo"""
+        all_packages = yield self._packagecloud_client.packages(
+            token=TOKEN, account=ACCOUNT, repo=repo).http_get()
+        raise gen.Return(all_packages)
+
+    @gen.coroutine
+    def _delete(self, packages_to_delete, repo, older_than=0,
+                number_to_keep=0):
+        """Generic packagecloud delete method, optionally supporting deleting
+        old packages by date."""
```
diranged: Write a huge and complete docstring here that describes the behavior very clearly. Describe the uses of `older_than`, `number_to_keep`, and how those interoperate with each other.

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,358 @@
...
+
+        # Extract a list of unique package names
+        package_names = [package['name'] for package in all_packages]
+        package_names = set(package_names)
+
+        pattern = re.compile(packages_to_delete)
+
+        # Loop through each unique package
+        for name in package_names:
+            if not pattern.match(name):
```
diranged: Rather than looping like this and using `continue` to skip packages we're not deleting, instead create the list of packages to delete ahead of time. Something like:

```
deletable_packages = [package['name'] in all_packages if pattern.match(package['name'])]

for package in deletable_packages:
  ...
```

diranged: I should note, this makes it easier to support `dry` mode and easily print out to a user:
```
"Would have deleted packages: %s" % packages
```

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,358 @@
...
+
+            packages.sort(key=lambda x: x.get('created_at'), reverse=True)
+
+            # Delete individual packages if they meet our criteria
+            number_in_repo = len(packages)
+            for package in packages:
+                log.debug('Inspecting package %s' % package)
+
+                if number_in_repo > number_to_keep:
+                    if not older_than:
```
diranged: This is not necessary.. Just

```
if older_than:
 ...
```

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,358 @@
...
+
+        pattern = re.compile(packages_to_delete)
+
+        # Loop through each unique package
+        for name in package_names:
+            if not pattern.match(name):
+                continue
+
+            # Build a sorted list of all versions of a package
+            packages = [{
```
diranged: `packages` is a confusing variable name here (since you already have `all_packages`, `deleted_packages`, and `package_names` defined above). How about `versions`?

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,358 @@
...
+        all_packages = yield self._get_all_packages(repo=repo)
+
+        # Extract a list of unique package names
+        package_names = [package['name'] for package in all_packages]
+        package_names = set(package_names)
+
+        pattern = re.compile(packages_to_delete)
+
+        # Loop through each unique package
+        for name in package_names:
```
diranged: To improve readability, I think the entire bit of logic inside this `for` loop should be pushed into its own method. This allows the code to be more readable:

```
# Fire off the package purges into the ioloop all at once
for package in deletable_packages:
  deletions.append(self._delete(package, repo, older_than, number_to_keep)

# Now yield on all of them simultaneously
yield deletions
```

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,358 @@
...
+
+    @gen.coroutine
+    def _get_all_packages(self, repo):
+        """Simple method for fetching a dictionary of all packages in a repo"""
+        all_packages = yield self._packagecloud_client.packages(
+            token=TOKEN, account=ACCOUNT, repo=repo).http_get()
+        raise gen.Return(all_packages)
+
+    @gen.coroutine
+    def _delete(self, packages_to_delete, repo, older_than=0,
```
diranged: Per my comment further below, you may want to rename `_delete` to `_delete_many`, and then move the `for ...` code block into `_delete()`.

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,358 @@
...
+
+            # Delete individual packages if they meet our criteria
+            number_in_repo = len(packages)
+            for package in packages:
+                log.debug('Inspecting package %s' % package)
+
+                if number_in_repo > number_to_keep:
+                    if not older_than:
+                        pass
+                    else:
```
diranged: This nested if/else/if/else bit is nasty and hard to understand. This should be something much simpler to read:

```
# If we've hit our limit of packages to keep, bail!
if number_to_keep <= number_in_repo:
  return

# If the package is younger than our minimum deletable package age, bail!
actual_package_age = (datetime.datetime.now() - package['created_at'])
min_package_age = datetime.timedelta(seconds=older_than)
if package_age <= min_package_age:
  return
```

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,358 @@
...
+                           datetime.timedelta(seconds=older_than):
+                            pass
+                        else:
+                            continue
+
+                    msg = '%s/%s/%s' % (
+                        repo,
+                        package['distro_version'],
+                        package['filename'])
+                    if self._dry:
```
diranged: Make this more pythonic ..

```
if self.dry:
  log.info('Would have deleted %s' % msg)
  return

log.info('Deleting ...')

```

kingpin/actors/packagecloud.py
cmclaughlin: Done

kingpin/actors/packagecloud.py
cmclaughlin: Done

kingpin/actors/packagecloud.py
cmclaughlin: Done, I called it extracted_packages

kingpin/actors/packagecloud.py
cmclaughlin: I took your suggestion - doing the pattern match and pulling out a list of packages to delete ahead of time, that way the set of packages to loop through is smaller.  However, it's a little more complicated than just saying "Would have deleted packages" with that info. Because the user may opt to keep a certain number of packages or delete by date and dryrun should respect that.

kingpin/actors/packagecloud.py
cmclaughlin: I don't understand what you're suggesting here.  Moving some code into a method is easy enough, but it looks like you're suggesting recursion too.  Please clarify.

cmclaughlin:
Addressed most of the feedback, not not all (there's a lot).  Tests are failing currently, but I'll have to come back to this.

cmclaughlin:
Just pushed another changeset that hopefully addresses some of Matt's comments.  I agree the _delete() method was too long and hard to read.  I just cleaned that up, but this still needs some more work, tests, etc.

cmclaughlin:
@diranged please take another look when you're back.

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,400 @@
...
+
+        # Loop through each unique package to delete
+        for name in packages_list_to_delete:
+            package_versions = yield self._get_package_versions(
+                name, all_packages)
+
+            # Delete individual packages if they meet our criteria
+            number_in_repo = len(package_versions)
+            for package in package_versions:
+                if number_in_repo > number_to_keep:
```
diranged: Reverse this... lets you out-dent lines 177-200.

```
if len(package_versions) <= number_to_keep:
  raise gen.Return()

if older_than:
...
```

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,400 @@
...
+                    if older_than:
+                        if datetime.datetime.now() - package['created_at'] > \
+                           datetime.timedelta(seconds=older_than):
+                            pass
```
diranged: Again this has some extra lines ... `pass` does nothing (https://docs.python.org/2/tutorial/controlflow.html#pass-statements). Instead:

```
if older_than and (datetime.datetime.now() - package['created_at'] <= datetime.timedelta(seconds=older_than)):
  continue
```

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,400 @@
...
+                    number_in_repo = number_in_repo - 1
+                    # Keeping track of removed packages for return value
+                    deleted_packages.append(package)
+
+        raise gen.Return(deleted_packages)
+
+
+class Delete(PackagecloudBase):
+
+    """packagecloud Delete Actor Class
```
diranged: Clean up the doc string here .. .the docs will actually be generated from this string. Consider something like:

```
Deletes old package versions from a PackageCloud repo.

... write up a full description of the behavior here. How the logic works with `number_to_keep` and `packages_to_delete`, etc.
```

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,400 @@
...
+        """Deletes all packages that match the `packages_to_delete` pattern"""
+        yield self._delete(
+            packages_to_delete=self.option('packages_to_delete'),
+            number_to_keep=self.option('number_to_keep'),
+            repo=self.option('repo'))
+
+
+class DeleteByDate(PackagecloudBase):
+
+    """packagecloud DeleteByDate Actor Class
```
diranged: Same comment as line 207

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,400 @@
...
+        yield self._delete(
+            packages_to_delete=self.option('packages_to_delete'),
+            number_to_keep=self.option('number_to_keep'),
+            older_than=self.option('older_than'),
+            repo=self.option('repo'))
+
+
+class WaitForPackage(PackagecloudBase):
+
+    """packagecloud WaitForPackage Actor
```
diranged: Same comment as line 207

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,400 @@
...
+        Args:
+            repo: name of the repo to search
+            name: Name of the package to search for as a regex
+            version: Version of the package to search for as a regex
+
+        Returns:
+            A list of the packages that were found
+        """
+
+        all_packages = yield self._packagecloud_client.packages(
```
diranged: Leverage `self._get_all_packages()`?

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,400 @@
...
+            matched_packages = yield self._search(
+                repo=self.option('repo'),
+                name=self.option('name'),
+                version=self.option('version'))
+
+            if len(matched_packages) > 0:
+                log.info('Found it!')
+                raise gen.Return()
+
+            yield gen.sleep(10)
```
diranged: The `10` should be an parameter I think.

diranged:
Looking great so far... couple of code comments inline. Additionally, your unit tests don't cover every scenario yet:

```
kingpin.actors.base                        183      2    99%   530, 532
kingpin.actors.packagecloud                 86     11    87%   182, 187, 251, 310, 387-400
```

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,400 @@
...
+            package_versions = yield self._get_package_versions(
+                name, all_packages)
+
+            # Delete individual packages if they meet our criteria
+            number_in_repo = len(package_versions)
+            for package in package_versions:
+                if number_in_repo > number_to_keep:
+                    if older_than:
+                        if datetime.datetime.now() - package['created_at'] > \
+                           datetime.timedelta(seconds=older_than):
```
siminm: I would create named values for these. What is `datetime.datetime.now() - package['created_at']` ? package_age would be great. Then the second part is allowed_age ? So your final if would say `if package_age > allowed_age` and that's very human readable

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,400 @@
...
+        'packages_to_delete': (
+            str, REQUIRED,
+            'Regex of packages to delete, e.g. pkg1|pkg2'),
+        'repo': (str, REQUIRED,
+                 'Which packagecloud repo to delete from')
+    }
+
+    @gen.coroutine
+    def _execute(self):
+        """Deletes all packages that match the `packages_to_delete` pattern and
```
siminm: Move this description to the Class. I don't think the _execute method needs a description. It's always "do the actual work for this class"

cmclaughlin:
Excellent review gents.  I still need to add a couple more tests for full coverage.

cmclaughlin:
@diranged this now has full coverage

siminm:
:+1:

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,416 @@
...
+      }
+    }
+
+    """
+
+    all_options = {
+        'name': (
+            str, REQUIRED, 'Name of the package to search for as a regex'),
+        'version': (
+            str, '.*', 'Version of the package to search for as a regex'),
```
diranged: Is there any valid regex checking here? What happens if the user types in something thats not regex?

cmclaughlin: Good idea, just added that.

diranged:
@siminm, check out my patches. Unit tests all pass now (had to re-org a bunch of data because of the bug on `reverse=True`).

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,486 @@
...
+                package['created_at'], '%Y-%m-%dT%H:%M:%S.%fZ'),
+            'distro_version': package['distro_version'],
+            'filename': package['package_html_url'].split('/')[-1],
+            'name': package['name']
+        } for package in all_packages if package['name'] == name]
+
+        versions.sort(key=lambda x: x.get('created_at'), reverse=False)
+        return versions
+
+    def _get_packages_list_to_delete(self, packages_to_delete, all_packages):
```
siminm: Were you thinking about renaming `packages_to_delete` ?

siminm: In fact, this code doesn't deal with any deletion at all. I'd change the method name to something like `_filter_packages(self, packages, regex)`

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,486 @@
...
+        try:
+            re.compile(self.option('packages_to_delete'))
+        except re.error:
+            raise exceptions.InvalidOptions(
+                'packages_to_delete is an invalid regex')
+
+    @gen.coroutine
+    def _execute(self):
+        """Deletes all packages that match the `packages_to_delete` pattern"""
+        yield self._delete(
```
siminm: Unconventional to invoke just 1 method within `_execute`. This method should find the packages needed for deletion, and then check the `_dry` setting.

kingpin/actors/packagecloud.py
```diff
@@ -0,0 +1,486 @@
...
+
+            matched_packages = yield self._search(
+                repo=self.option('repo'),
+                name=self.option('name'),
+                version=self.option('version'))
+
+            if len(matched_packages) > 0:
+                self.log.info('Found it!')
+                raise gen.Return()
+
```
siminm: Add a debug line saying "Not found, sleeping for %s"

siminm:
@diranged address the comments and :shipit:

kingpin/actors/packagecloud.py
diranged: Given the complexity of the loop, I am OK with it being the way it is now. If we want to do what you suggest, it means that we have to create several more methods and handle passing a ton of data around.

kingpin/actors/packagecloud.py
diranged: Done

diranged:
@siminm, comments addressed... merging.

PR #278, opened 2015-07-08
Add Concurrency Limit to Execute Actor

siminm:
Haven't written tests yet because I"m not sure this is the best approach.
Quite a bit of change here -- there's a 100% different flow of work if concurrency limit is added.


siminm:
cc: @diranged all clean and tested now. Reworked the logic even further to allow multiple-arrays to be limited in concurrency through instance count. So as an example:

two arrays A and B with A having 1 instance and B having 5
Concurrency set to 2

This will start 1 in A and 1 in B, and as soon as either one of those tasks finish, this will proceed to the second instance in B.

siminm:
@diranged @cmclaughlin please review. It would help not slam our Puppet masters when running noop on 100s of machines

siminm:
Fixed the issue @cmclaughlin found.

cmclaughlin:
:+1: :+1:

PR #282, opened 2015-08-20
Create rightscale deployment Destroy actor

siminm:
```
$ nosetests kingpin/actors/rightscale/test/integration_deployment.py --include=integration
....
----------------------------------------------------------------------
Ran 4 tests in 5.862s

OK
```

### DIFF
--- a/kingpin/actors/rightscale/deployment.py
+++ b/kingpin/actors/rightscale/deployment.py
@@ -95,6 +95,11 @@ class Create(DeploymentBaseActor):
     @gen.coroutine
     def _execute(self):
 
+        dep = yield self._find_deployment(self.option('name'))
+        if dep:
+            raise exceptions.InvalidOptions(
+                'Deployment "%s" already exists.' % self.option('name'))
+
         params = {'name': self.option('name'),
                   'description': self.option('description')}
 
@@ -108,14 +113,10 @@ class Create(DeploymentBaseActor):
             self.log.debug('Deployment params: %s' % params)
             raise gen.Return()
 
-        try:
-            yield self._client.create_resource(
-                self._client._client.deployments, params)
-        except Exception as e:
-            self.log.debug('Encountered error: %s' % e)
-            raise exceptions.RecoverableActorFailure(
-                'Could not create deployment %s. Does it exist already?' %
-                self.option('name'))
+        self.log.info('Creating deployment %s' % self.option('name'))
+
+        yield self._client.create_resource(
+            self._client._client.deployments, params)
 
 
 class Destroy(DeploymentBaseActor):
@@ -139,6 +140,10 @@ class Destroy(DeploymentBaseActor):
     def _execute(self):
 
         dep = yield self._find_deployment(self.option('name'))
+        if not dep:
+            raise exceptions.InvalidOptions(
+                'Deployment "%s" does not exist.' % self.option('name'))
+
         info = (yield self._client.show(dep.self)).soul
 
         if self._dry:
--- a/kingpin/actors/rightscale/deployment.py
+++ b/kingpin/actors/rightscale/deployment.py
@@ -37,6 +37,15 @@ class DeploymentBaseActor(base.RightScaleBaseActor):
 
     """Abstract Deployment Actor that provides some utility methods."""
 
+    @gen.coroutine
+    def _find_deployment(self, name):
+
+        dep = yield self._client.find_by_name_and_keys(
+            collection=self._client._client.deployments,
+            name=name)
+
+        raise gen.Return(dep)
+
 
 class Create(DeploymentBaseActor):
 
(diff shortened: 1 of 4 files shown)


kingpin/actors/rightscale/deployment.py
```diff
@@ -37,6 +37,15 @@ class DeploymentBaseActor(base.RightScaleBaseActor):
 
     """Abstract Deployment Actor that provides some utility methods."""
 
+    @gen.coroutine
```
diranged: retry decorator?

kingpin/actors/rightscale/deployment.py
```diff
@@ -98,13 +107,43 @@ def _execute(self):
             self.log.info('Would create a deployment %s' % self.option('name'))
             self.log.debug('Deployment params: %s' % params)
             raise gen.Return()
-        self.log.info('Deployment params: %s' % params)
 
         try:
             yield self._client.create_resource(
```
diranged: Is this wrapped in a retry decorator inside the API class?

kingpin/actors/rightscale/deployment.py
```diff
@@ -98,13 +107,43 @@ def _execute(self):
...
+    """
+
+    all_options = {
+        'name': (str, REQUIRED, 'The name of the deployment to be deleted.'),
+    }
+
+    @gen.coroutine
+    def _execute(self):
+
+        dep = yield self._find_deployment(self.option('name'))
```
diranged: What is raised if this deployment doesn't exist?

kingpin/actors/rightscale/test/integration_deployment.py
```diff
@@ -0,0 +1,51 @@
...
+    integration = True
+
+    def setUp(self, *args, **kwargs):
+        super(IntegrationDeployment, self).setUp(*args, **kwargs)
+        self.deployment_name = 'Kingpin Deployment Integration Test'
+
+    @attr('integration', 'dry')
+    @testing.gen_test()
+    def integration_01a_create_deployment_dry(self):
+        actor = deployment.Create('Integ. Test',
```
diranged: Make the deployment name more dynamic -- so its unique per test. For example, look at the way we name the ServerArrays in those integration tests.

kingpin/actors/rightscale/deployment.py
```diff
@@ -98,13 +107,43 @@ def _execute(self):
             self.log.info('Would create a deployment %s' % self.option('name'))
             self.log.debug('Deployment params: %s' % params)
             raise gen.Return()
-        self.log.info('Deployment params: %s' % params)
 
         try:
             yield self._client.create_resource(
                 self._client._client.deployments, params)
         except Exception as e:
```
diranged: Wait -- I missed this before. A blanket Exception catch? No no. Please fix.

kingpin/actors/rightscale/deployment.py
siminm: this really just calls `find_by_name_nad_keys()` the api module has a retry decorator there

kingpin/actors/rightscale/deployment.py
siminm: yes.

kingpin/actors/rightscale/deployment.py
siminm: I'd have to catch `requests.http.exceptions` and would import `requests` just for that... that's ok?

Any way to configure `rightscale.settings.RETRYING_SETTINGS` similar to our tornado retry? To raise a specified exception for a matched string?

nextdoor-ops-ro: Go ahead and import the requests library,... its getting loaded up anyways.
:)

Matt Wise
Sr. Systems Architect
Nextdoor.com

siminm:
Addressed exception raising/handling by reworking that logic. Explicitly searching for the deployment before destroying it. This helps with a dry run also, and no need to import `requests` just to catch the exception

kingpin/actors/rightscale/deployment.py
```diff
@@ -98,13 +112,43 @@ def _execute(self):
...
+                'Deployment "%s" does not exist.' % self.option('name'))
+
+        info = (yield self._client.show(dep.self)).soul
+
+        if self._dry:
+            self.log.info('Would delete deployment %s' % info['name'])
+            raise gen.Return()
+
+        self.log.info('Deleting deployment %s' % info['name'])
+        yield self._client.destroy_resource(dep)
```
diranged: Something just occurred to me as I was reading this -- what happens if we try to delete a deployment with servers in it? Is an error returned back to us? Can you try that out real quick?

diranged:
:shipit: ... one quick question in there, but it can be dealt with separately if you want.

siminm:
Hrm, yes it raises RS's catch-all `422` :( Since the destroy actor is there only for completion of integration tests I'd rather deal with it later

PR #283, opened 2015-08-25
Create rightscale.deployment.Clone actor

mikhail:
- [x] Add docs
- [x] Add unit tests
- [x] Add integration tests

### DIFF
--- a/kingpin/actors/rightscale/deployment.py
+++ b/kingpin/actors/rightscale/deployment.py
@@ -199,7 +199,7 @@ class Clone(DeploymentBaseActor):
                 arrays = dep.server_arrays.show()
                 self.log.info('Found %s arrays' % len(arrays))
                 for a in arrays:
-                    self.log.info('Would delete %s' % s.soul['name'])
+                    self.log.info('Would delete %s' % a.soul['name'])
 
             raise gen.Return()
 
@@ -219,7 +219,7 @@ class Clone(DeploymentBaseActor):
             arrays = newdep.server_arrays.show()
             self.log.info('Found %s arrays' % len(arrays))
             for a in arrays:
-                self.log.info('Deleting %s' % s.soul['name'])
+                self.log.info('Deleting %s' % a.soul['name'])
                 yield self._client.destroy_resource(a)
 
             self.log.info('Done deleting servers and arrays.')
--- a/kingpin/actors/rightscale/deployment.py
+++ b/kingpin/actors/rightscale/deployment.py
@@ -51,9 +51,6 @@ class Create(DeploymentBaseActor):
 
     """Creates a RightScale deployment.
 
-    Options match the documentation in RightScale:
-    http://reference.rightscale.com/api1.5/resources/ResourceDeployments.html
-
     **Options**
 
     :name:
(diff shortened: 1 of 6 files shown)


kingpin/actors/rightscale/deployment.py
```diff
@@ -119,13 +116,106 @@ def _execute(self):
...
+            self.log.info('Found %s servers' % len(servers))
+            for s in servers:
+                self.log.info('Deleting %s' % s.soul['name'])
+                yield self._client.destroy_resource(s)
+
+            arrays = newdep.server_arrays.show()
+            self.log.info('Found %s arrays' % len(arrays))
+            for a in arrays:
+                self.log.info('Deleting %s' % s.soul['name'])
+                yield self._client.destroy_resource(a)
```
diranged: These two calls are interesting -- I like the idea, but I'm concerned about them. If there are servers running, you won't be able to delete the Server or ServerArray. Maybe we'll need to create code that can handle that scenario?  Maybe we throw a warning or even an error here if thats the case?

siminm: Impossible -- this is a freshly cloned deployment and the servers can't be running already.

nextdoor-ops-ro: Not impossible .. improbable. :)

Matt Wise
Sr. Systems Architect
Nextdoor.com

siminm: Just confirmed that it's actually impossible. Servers are created as definitions (not launched) and arrays are cloned in _disabled_ state

siminm:
@diranged ready for review

kingpin/actors/rightscale/deployment.py
```diff
@@ -119,13 +116,119 @@ def _execute(self):
...
+       (*optional*)
+
+    :server_tag_scope:
+       The routing scope for tags for servers in the deployment.
+       Can be 'deployment' or 'account'
+       (*optional*, default: deployment)
+    """
+
+    all_options = {
+        'clone': (str, REQUIRED, 'The name of the deployment to clone.'),
```
diranged: Can you rename `clone` to `source` and `name` to `dest`? This follows our other `.Clone` actors more closely.

kingpin/actors/rightscale/deployment.py
```diff
@@ -119,13 +116,119 @@ def _execute(self):
...
+    @gen.coroutine
+    def _execute(self):
+
+        dep = yield self._find_deployment(self.option('clone'))
+        if not dep:
+            raise exceptions.InvalidOptions(
+                'Deployment "%s" does not exist.' % self.option('clone'))
+        self.log.info('Found "%s" to clone.' % self.option('clone'))
+
+        newdep = yield self._find_deployment(self.option('name'))
```
diranged: I know its a little bit silly, but you can group lines 170 and 176 so that the calls happen asynchronously. :)

kingpin/actors/rightscale/deployment.py
siminm: I wanted to keep `name` because that's the RS API name for the destination name.  http://reference.rightscale.com/api1.5/resources/ResourceDeployments.html ... thoughts?

kingpin/actors/rightscale/deployment.py
```diff
@@ -119,13 +116,119 @@ def _execute(self):
...
+            raise exceptions.InvalidOptions(
+                'Deployment "%s" already exists.' % self.option('name'))
+
+        optional_params = {'name', 'description', 'server_tag_scope'}
+        params = {}
+        for p in optional_params:
+            if self.option(p):
+                params[p] = self.option(p)
+
+        params = self._generate_rightscale_params('deployment', params)
```
diranged: You may want to move lines 182-188 into the `__init__` method so that we can catch errors faster.

kingpin/actors/rightscale/deployment.py
```diff
@@ -119,13 +116,119 @@ def _execute(self):
...
+                self.log.info('Found %s arrays' % len(arrays))
+                for a in arrays:
+                    self.log.info('Would delete %s' % a.soul['name'])
+
+            raise gen.Return()
+
+        self.log.info('Creating deployment %s' % self.option('name'))
+
+        dep_id = self._client.get_res_id(dep)
+        newdep = self._client._client.deployments.clone(
```
diranged: Not yielding?

kingpin/actors/rightscale/deployment.py
```diff
@@ -119,13 +116,119 @@ def _execute(self):
...
+                arrays = dep.server_arrays.show()
+                self.log.info('Found %s arrays' % len(arrays))
+                for a in arrays:
+                    self.log.info('Would delete %s' % a.soul['name'])
+
+            raise gen.Return()
+
+        self.log.info('Creating deployment %s' % self.option('name'))
+
+        dep_id = self._client.get_res_id(dep)
```
diranged: Not yielding?

kingpin/actors/rightscale/deployment.py
```diff
@@ -119,13 +116,119 @@ def _execute(self):
...
+        newdep = self._client._client.deployments.clone(
+            res_id=dep_id, params=params)
+
+        if self.option('delete_servers'):
+            servers = newdep.servers.show()
```
diranged: Not yielding?

kingpin/actors/rightscale/deployment.py
```diff
@@ -119,13 +116,119 @@ def _execute(self):
...
+            self.log.info('Found %s servers' % len(servers))
+            for s in servers:
+                self.log.info('Deleting %s' % s.soul['name'])
+                yield self._client.destroy_resource(s)
+
+            arrays = newdep.server_arrays.show()
```
diranged: Not yielding?

kingpin/actors/rightscale/deployment.py
siminm: This is just a string-parsing function (api.py:114) -- no generator here (you wrote it!) so no yielding.

kingpin/actors/rightscale/deployment.py
siminm: Note this is a _client._client -- same logic we use for cloning server arrays (api.py:305)

kingpin/actors/rightscale/deployment.py
siminm: `newdep` is a raw RightScale resource. I'll leverage the `api.show()` method here instead of using the raw call

diranged:
@siminm  you want to check on my comments and submit a new commit?

PR #287, opened 2015-10-02
Use a 120s timeout when interacting with PackageCloud

diranged:
Without this, Tornado defaults to a `20s` timeout and then throws a 599 error on its own. The PackageCloud API sometimes takes a while to respond if you're asking for a ton of data.


diranged:
Reviewers: @siminm, @nrvale0

siminm:
:+1:

diranged:
Tested it and it worked... merging.

PR #288, opened 2015-10-02
Fix doc typo

siminm:
cc: @nrvale0


siminm:
Approved in person with @nrvale0

PR #289, opened 2015-10-21
Initial support for creating and destroying RightScale MCIs.

diranged:
No tests yet.

### DIFF
--- a/kingpin/actors/rightscale/api.py
+++ b/kingpin/actors/rightscale/api.py
@@ -227,7 +227,7 @@ class RightScale(object):
     @concurrent.run_on_executor
     @sync_retry(**settings.RETRYING_SETTINGS)
     @utils.exception_logger
-    def find_by_name_and_keys(self, collection, name, exact=True, **kwargs):
+    def find_by_name_and_keys(self, collection, exact=True, **kwargs):
         """Search for a RightScale resource by name, and optional keys.
 
         This code is blatently stolen from rightscale.util.find_by_name and
@@ -238,7 +238,6 @@ class RightScale(object):
 
         Args:
             collection: RightScale.<xxx> resource object
-            name: Name of the resource to search for
             exact: If True, returns the first match. If False, returns a list
                 of all returned resources.
             **kwargs: Any additional keys-and-values to use in the search.
@@ -247,18 +246,21 @@ class RightScale(object):
             RightScale Resource Objects
         """
         filter_keys = []
-        kwargs['name'] = name
         for key, val in kwargs.items():
             filter_keys.append('%s==%s' % (key, val))
-        params = {'filter[]': filter_keys}
+        params = {'filter[]': sorted(filter_keys)}
 
         found = collection.index(params=params)
         if not exact and len(found) > 0:
             return found
 
-        for f in found:
-            if f.soul['name'] == name:
-                return f
+        if len(found) < 1:
+            return None
+
+        if len(found) == 1:
+            return found[0]
+
+        return found
 
     @concurrent.run_on_executor
     @sync_retry(**settings.RETRYING_SETTINGS)
--- a/kingpin/actors/rightscale/api.py
+++ b/kingpin/actors/rightscale/api.py
@@ -243,7 +243,7 @@ class RightScale(object):
             **kwargs: Any additional keys-and-values to use in the search.
 
         Returns:
-            RightScale Resource Objects
+            One RightScale Resource Object or a List of objects.
         """
         filter_keys = []
         for key, val in kwargs.items():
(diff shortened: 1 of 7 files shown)


kingpin/actors/rightscale/api.py
```diff
@@ -227,7 +227,7 @@ def find_right_script(self, name):
     @concurrent.run_on_executor
     @sync_retry(**settings.RETRYING_SETTINGS)
     @utils.exception_logger
-    def find_by_name_and_keys(self, collection, name, exact=True, **kwargs):
+    def find_by_name_and_keys(self, collection, exact=True, **kwargs):
```
siminm: Why are you removing name? I have code that's leveraging that

nextdoor-ops-ro: Actually its not -- name still works just fine, I just removed the
'requirement' that you supply it. There are cases where you want to search
on something else (like a resource_uid). However, I may nee to sanity check
our code and make sure we're always passing in `name=<myname>` rather than
just `<myname>`.

Matt Wise
Sr. Systems Architect
Nextdoor.com

kingpin/actors/rightscale/mci.py
```diff
@@ -0,0 +1,227 @@
...
+            raise gen.Return()
+
+        # Ok, lets create this thing!
+        self.log.info('Creating MCI %s' % self.option('name'))
+        mci = yield self._client.create_resource(
+            self._client._client.multi_cloud_images,
+            mci_params)
+
+        # Now add each of the image descriptions to the mci
+        for setting in mci_settings_params:
```
siminm: Parallel these?

siminm:
Overall looks good. Some obvious copy/paste errors, but highlevel is :+1:

kingpin/actors/rightscale/mci.py
```diff
@@ -0,0 +1,295 @@
...
+
+    """Abstract MCI Actor that provides some utility methods."""
+
+
+class Create(MCIBaseActor):
+
+    """Creates a RightScale Multi Cloud Image.
+
+    Options match the documentation in RightScale:
+    http://reference.rightscale.com/api1.5/resources/ResourceDeployments.html
```
siminm: Nope.

kingpin/actors/rightscale/api.py
```diff
@@ -247,18 +246,21 @@ def find_by_name_and_keys(self, collection, name, exact=True, **kwargs):
...
         if not exact and len(found) > 0:
             return found
 
-        for f in found:
-            if f.soul['name'] == name:
-                return f
+        if len(found) < 1:
+            return None
+
+        if len(found) == 1:
```
siminm: This is pretty bad... There shouldn't be any guesswork at to what _type_ of data gets returned. None, Object, or List. I strongly recommend always returning a list. List of 0, list of 1 or list of many...

kingpin/actors/rightscale/mci.py
```diff
@@ -0,0 +1,295 @@
...
+              ]
+          }
+        }
+
+    """
+
+    all_options = {
+        'name': (str, REQUIRED, 'The name of the MCI to be created.'),
+        'description': (
+            str, '',
```
siminm: Maybe default to "Created via Kingpin" ? Kinda like "Sent from BlackBerry"

kingpin/actors/rightscale/mci.py
```diff
@@ -0,0 +1,295 @@
...
+            # Make sure that the required options were passed in
+            for required in required_image_options:
+                if required not in image.keys():
+                    raise exceptions.InvalidOptions(
+                        'Missing option "%s" in Image %s' %
+                        (required, image))
+
+    @gen.coroutine
+    def _get_image_def(self, description):
+        """Returns a fully populated set of Multi Cloud Image settings.
```
siminm: nitpicky, but the docstring here is inconsistent about using `set` vs `array of tuples` when describing the return data.

kingpin/actors/rightscale/mci.py
```diff
@@ -0,0 +1,295 @@
...
+            description: A dictionary with the keys: cloud, image,
+            instance_type, user_data
+
+        Returns:
+            A RightScale-formatted array of tuples.
+        """
+
+        # Get our cloud object first -- its required so that we can search for
+        # the image/ramdisk/etc hrefs.
+        cloud = yield self._client.find_by_name_and_keys(
```
siminm: Each time this is invoked it'll get a new cloud name, right? Just thinking if any caching could optimize this.

siminm:
Two big things - return type and wrong doc string. Address those and merge it

diranged:
Merging -- the PyPY tests fail because its so slow that our timer-based tests fail. This is a separate issue that we need to look into.

PR #290, opened 2015-10-21
Depend on the newest RightScale library patch.

diranged:
@siminm, can you review this and the underlying python-rightscale change that i made? if you think its decent enough, we can at least merge this into our code base..


siminm:
Hard to compare since I don't know what the diff is. I clicked around on the recent commits and they're all very reasonable. If this works - ::shipit::

PR #292, opened 2015-10-22
Make rightscale.api.find_by_name_and_keys() return only a list or a s…

diranged:
…ingle object.

Instead of returning None, return an empty list. This is more consistent
with the way the original stolen method works anyways.


kingpin/actors/rightscale/test/test_api.py
```diff
@@ -116,7 +116,7 @@ def test_find_by_name_and_keys(self):
         ret = yield self.client.find_by_name_and_keys(
             collection=mock.MagicMock(), exact=True,
             name='FakeResource', href='/123')
-        self.assertEquals(ret, None)
```
siminm: :-/ Not what I meant. If `exact=True` is provided that means it should be exactly 1 object, right? So it probably should return an object or raise an exception. 

If `exact=False` then it's a search, and therefore always a list.

PR #296, opened 2015-10-28
Add context-file

siminm:


siminm:
cc: @diranged

kingpin/actors/group.py
```diff
@@ -50,6 +51,7 @@ class BaseGroupActor(base.BaseActor):
 
     all_options = {
         'contexts': (list, [], "List of contextual hashes."),
+        'context-file': (str, '', "A JSON file with a list of contexts."),
```
diranged: So ... I think I'd rather re-use `contexts` and simply have the code check for a `str` or a `list` and behave differently. Thoughts?

kingpin/actors/group.py
```diff
@@ -89,11 +91,21 @@ def _build_actions(self):
...
         """
-        if not self.option('contexts'):
+        any_contexts = self.option('contexts') or self.option('context-file')
+        if not any_contexts:
             return self._build_action_group(self._init_context)
 
+        if self.option('contexts'):
+            context_data = self.option('contexts')
+            if self.option('context-file'):
+                raise exceptions.InvalidOptions(
```
diranged: By re-using the word `contexts` above, you remove this complication. It clearly forces the user to do one or the other.

diranged:
Fundamentally this is great -- I agree with it whole-heartedly. However, I think we should re-use the `contexts` option and just let it intelligently handle a `str` file reference or a `list` of dicts.

Beyond that, the docs need clean up (kingpin is mispelled, one of the doc lines is too long, etc).

kingpin/actors/group.py
siminm: I don't think I like that. I'll try to formulate exactly why, but in general I'm not a fan of overloading a variable.

siminm:
Ready for review. @diranged @bemsha

docs/basicuse.rst
```diff
@@ -338,6 +338,45 @@ ability to define usable tokens, but any actor can then reference these tokens.
     2015-01-14 15:02:22,165 INFO      [DRY: Notify Engineering] Sending message "Hey room .. I'm done with the release. Get back to work" to Hipchat room "Engineering"
     2015-01-14 15:02:22,239 INFO      [DRY: Notify Cust Service] Sending message "Hey room .. I'm done with the release. Have a nice day" to Hipchat room "Cust Service"
 
+Contextual tokens stored in separate file
+'''''''''''''''''''''''''''''''''''''''''
+
+When multiple kingping json files need to leverage the same context for different purposes it is useful to put the contexts into a stand alone file and then reference that file. Since `%TOKENS%` won't be parsed inside of the context file they need to be explicitly passed into it.
```
diranged: Break out into multiple lines.

diranged: `Kingpin` ... not `kingping`

diranged: capitalize `JSON`

diranged: Re-word: "Since `%TOKENS%` won't be parsed inside of the context file they need to be explicitly passed into it."
Suggest: "Context files support [tokens](#<link to env token section above>) just like our [Macros](#<link to macro actor>) do -- you must pass in the `%TOKEN%` and its `value`. See below for examples."

kingpin/actors/group.py
```diff
@@ -71,6 +73,12 @@ def __init__(self, *args, **kwargs):
         """
         super(BaseGroupActor, self).__init__(*args, **kwargs)
 
+        if type(self.option('contexts')) == dict:
+            try:
+                open(self.option('contexts')['file'])
+            except IOError as e:
```
diranged: If someone passes in a `dict` that doesn't have the `file` key, this will throw a different exception. Test for that. :)

kingpin/actors/group.py
```diff
@@ -89,13 +97,26 @@ def _build_actions(self):
...
+        if not contexts:
             return self._build_action_group(self._init_context)
 
+        if type(contexts) == list:
+            context_data = self.option('contexts')
+        elif type(contexts) == dict:
+            context_string = open(contexts['file']).read()
+            context_string = kp_utils.populate_with_tokens(
+                string=context_string,
+                tokens=contexts['tokens'],
```
diranged: Check for `tokens` before calling it -- if its not there, this will fail.

kingpin/actors/group.py
```diff
@@ -89,13 +97,26 @@ def _build_actions(self):
...
+                strict=True)
+            context_data = demjson.decode(context_string)
+
         actions = []
-        for context in self.option('contexts'):
+        for context in context_data:
             combined_context = dict(self._init_context.items() +
                                     context.items())
+            self.log.debug('Init context: %s' % self._init_context.items())
```
diranged: Use different phrases here -- I know they match the variable names, but to the "caller" of this code they wouldn't be terribly useful.

docs/basicuse.rst
siminm: Done x 3.

kingpin/actors/group.py
siminm: done

kingpin/actors/group.py
siminm: Done

siminm:
@diranged addressed all your comments (without arguing!)

docs/basicuse.rst
```diff
@@ -338,6 +338,48 @@ ability to define usable tokens, but any actor can then reference these tokens.
     2015-01-14 15:02:22,165 INFO      [DRY: Notify Engineering] Sending message "Hey room .. I'm done with the release. Get back to work" to Hipchat room "Engineering"
     2015-01-14 15:02:22,239 INFO      [DRY: Notify Cust Service] Sending message "Hey room .. I'm done with the release. Have a nice day" to Hipchat room "Cust Service"
 
+Contextual tokens stored in separate file
+'''''''''''''''''''''''''''''''''''''''''
+
+When multiple Kingping JSON files need to leverage the same context for
```
diranged: Umm .. there's still a `g` here in Kingpin. :panda_face:

diranged:
Fix the `g` and then merge it. :+1:

PR #299, opened 2015-11-24
Allow server_array.Launch to not enable and not launch.

siminm:
cc: @diranged @bemsha


diranged:
Add a unit test that proves the behavior we expect...

siminm:
bump

diranged:
:+1:

PR #302, opened 2015-12-04
Don't sort rightscale params.

siminm:
cc: @diranged @bemsha

### DIFF
--- a/kingpin/actors/rightscale/test/test_base.py
+++ b/kingpin/actors/rightscale/test/test_base.py
@@ -130,39 +130,41 @@ class TestRightScaleBaseActor(testing.AsyncTestCase):
 
     def test_generate_rightscale_params_with_array(self):
         self.maxDiff = None
-        params = {'name': 'unittest-name',
-                  'status': 'enabled',
-                  'elasticity_params': {
-                      'schedule': [
-                          {'day': 'Sunday', 'max_count': 2,
-                           'min_count': 1, 'time': '07:00'},
-                          {'day': 'Monday', 'max_count': 2,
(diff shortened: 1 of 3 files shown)


kingpin/actors/rightscale/test/test_base.py
```diff
@@ -130,22 +130,16 @@ def test_generate_rightscale_params(self):
...
-                           'min_count': 1, 'time': '07:00'}
-                      ]
-                  }}
-        expected_params = [
-            ('server_array[status]', 'enabled'),
-            ('server_array[name]', 'unittest-name'),
+        schedule = [
+            {'day': 'Sunday', 'max_count': 2, 'min_count': 1, 'time': '07:00'},
+            {'day': 'Monday', 'max_count': 2, 'min_count': 1, 'time': '07:00'},
+            {'day': 'Tuesday', 'max_count': 2, 'min_count': 1, 'time': '07:00'}
```
bemsha: Doesn't this test assume that assigning to a dict changes the order in which items are returned? Wouldn't it be a stronger test if each dict were an OrderedDict with its items in a different order?

siminm: What is inside of this python code doesn't really matter because in reality the code is parsed from the JSON file, and cannot be specified as an OrderedDict. As far as I'm aware RightScale schedule is the only one that requires particular order of parameters.

siminm:
@diranged fixed the test to actually check for "groups of 4" as per rightscale documentation

diranged:
:+1: 

(damn, that `zip` code is ugly!)

PR #304, opened 2015-12-08
Add elb.UpdateCert actor

mikhail:
cc: @diranged @nrvale0

### DIFF
--- a/kingpin/actors/aws/elb.py
+++ b/kingpin/actors/aws/elb.py
@@ -66,6 +66,78 @@ class ELBBaseActor(base.AWSBaseActor):
         'region': (str, REQUIRED, 'AWS region (or zone) name, like us-west-2')
     }
 
+    @gen.coroutine
+    def _check_access(self, elb):
+        """Perform a dummy operation to check credential accesss.
+
+        Intended to be used in a dry run, this method attempts to perform an
+        invalid set_listener call and monitors the output of the error. If the
+        error is anything other than AccessDenied then the provided credentials
+        are sufficient and we do nothing.
+
+        Args:
+            elb: boto LoadBalancer object.
+        """
+        try:
+            # A blank ARN value should have code 'CertificateNotFound'
+            # We're only checking if credentials have sufficient access
+            yield self.thread(
+                elb.set_listener_SSL_certificate,
+                self.option('port'),
+                '')
+        except BotoServerError as e:
+            if e.error_code == 'AccessDenied':
+                raise exceptions.InvalidCredentials(e)
+
+    @gen.coroutine
+    def _use_cert(self, elb, arn, port):
+        """Assign an ssl cert to a given ELB.
+
+        Args:
+            elb: boto elb object.
+            arn: ARN for server certificate to use.
+        """
+
+        self.log.info('Setting ELB "%s" to use cert arn: %s' % (elb, arn))
+        try:
+            yield self.thread(
+                elb.set_listener_SSL_certificate, port, arn)
+        except BotoServerError as e:
+            raise exceptions.RecoverableActorFailure(
+                'Applying new SSL cert to %s failed: %s' % (elb, e))
+
+    @gen.coroutine
+    def _get_cert_arn(self, name):
+        """Return a server_certificate ARN.
+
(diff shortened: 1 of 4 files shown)


kingpin/actors/aws/elb.py
```diff
@@ -365,6 +365,150 @@ def _execute(self):
...
+      (str) Unique IAM certificate name, or ARN
+
+    **Example**
+
+    .. code-block:: json
+
+       { "actor": "aws.elb.SetCert",
+         "desc": "Run SetCert",
+         "options": {
+           "cert_name": "new-cert",
```
diranged: This example doesn't make sense? Copy and paste error?

kingpin/actors/aws/elb.py
```diff
@@ -365,6 +365,150 @@ def _execute(self):
...
+
+        self.log.info('Setting ELB "%s" to use cert arn: %s' % (elb, arn))
+        try:
+            yield self.thread(
+                elb.set_listener_SSL_certificate, port, arn)
+        except BotoServerError as e:
+            raise exceptions.RecoverableActorFailure(
+                'Applying new SSL cert to %s failed: %s' % (elb, e))
+
+    def _find_cert_port(self, elb, new_arn):
```
diranged: Little formatting and doc string issues here.. remove the extra line, document the code.

kingpin/actors/aws/elb.py
```diff
@@ -365,6 +365,150 @@ def _execute(self):
...
+        for lis in elb.listeners:
+            ssl_arn = lis[4]
+            port = lis[0]
+            if ssl_arn == new_arn:
+                return port
+
+        return False
+
+    @gen.coroutine
+    def _execute(self):
```
diranged: Just because we can, reformat this to operate fully asynchronously across all the ELBs at once?

diranged:
Basically looks good. I think you could move a few of those methods out into a `BaseCert` type class, but we can do that as we add more certificate management classes.

siminm:
bump

kingpin/actors/aws/elb.py
```diff
@@ -66,6 +66,78 @@ class ELBBaseActor(base.AWSBaseActor):
...
+
+        Args:
+            elb: boto LoadBalancer object.
+        """
+        try:
+            # A blank ARN value should have code 'CertificateNotFound'
+            # We're only checking if credentials have sufficient access
+            yield self.thread(
+                elb.set_listener_SSL_certificate,
+                self.option('port'),
```
diranged: The problem with this is that if you use the `_check_access()` method in any other AWS actor, it may not have the `port` option. Perhaps just fake the port out with something invalid?

kingpin/actors/aws/elb.py
```diff
@@ -362,7 +362,93 @@ def _execute(self):
             self.log.info('Would instruct %s to use %s' % (
                 self.option('name'), self.option('cert_name')))
         else:
-            yield self._use_cert(elb, cert_arn)
+            yield self._use_cert(elb, cert_arn, self.option('port'))
+
+
+class UpdateCert(ELBBaseActor):
```
diranged: I'm thinking that maybe this name should change -- `UpdateCertReferences`? I just find it a little vague as it is now. Thoughts?

kingpin/actors/aws/elb.py
```diff
@@ -362,7 +362,93 @@ def _execute(self):
...
+            False: if no listener is using specified certificate.
+        """
+
+        for lis in elb.listeners:
+            ssl_arn = lis[4]
+            port = lis[0]
+            if ssl_arn == cert_arn:
+                return port
+
+        return False
```
diranged: Just curious - why not `None`?

kingpin/actors/aws/elb.py
siminm: This is very ELB specific - it can't be a checker for anything other than ELB

kingpin/actors/aws/elb.py
siminm: No reason :-/ `None` is probably better - more pythonic this way.

PR #307, opened 2015-12-10
Take in Kingpin actor definition on the commandline

diranged:
@siminm, this is a wip .. but whatcha think so far? Basically I want us to be able to easily run simple actors on the commandline without generating JSON.

```
(.venv)Matts-MacBook:kingpin diranged$ ./kingpin/bin/deploy.py -c  --actor misc.Sleep -p warn_on_failure=true -o sleep=2
17:15:52   INFO      Rehearsing... Break a leg!
17:15:52   INFO      [DRY: Kingpin] Preparing actors from {'warn_on_failure': 'true', 'options': {'sleep': '2'}, 'actor': 'misc.Sleep', 'desc': 'Commandline Execution'}
17:15:52   INFO      Rehearsal OK! Performing!
17:15:52   INFO      [Kingpin] Preparing actors from {'warn_on_failure': 'true', 'options': {'sleep': '2'}, 'actor': 'misc.Sleep', 'desc': 'Commandline Execution'}
17:15:53   INFO      
17:15:53   WARNING   Lights, camera ... action!
17:15:53   INFO      
(.venv)Matts-MacBook:kingpin diranged$ 
```

Just an example..

### DIFF
--- a/kingpin/actors/misc.py
+++ b/kingpin/actors/misc.py
@@ -166,6 +166,12 @@ class Macro(base.BaseActor):
         open the local file and return a buffer to that file.
         """
 
+        if self.option('macro').startswith('{'):
+            buf = StringIO.StringIO()
+            buf.write(self.option('macro'))
+            buf.seek(0)
+            return buf
+
         remote = ('http://', 'https://')
         if self.option('macro').startswith(remote):
             client = httpclient.HTTPClient()
--- a/kingpin/actors/test/test_misc.py
+++ b/kingpin/actors/test/test_misc.py
@@ -38,6 +38,17 @@ class TestMacro(testing.AsyncTestCase):
             self.assertEquals(schema_validate.call_count, 1)
             self.assertEquals(actor.initial_actor, get_actor())
 
+    def test_init_raw_json(self):
+        json_script = ('{"desc": "",'
+                       ' "actor": "misc.Sleep",'
+                       ' "options": {"sleep": 0.1}}')
+        misc.Macro._get_config_from_json = mock.Mock()
+        misc.Macro._check_schema = mock.Mock()
+        with mock.patch('kingpin.actors.utils.get_actor'):
+            with mock.patch.object(httpclient.HTTPClient, 'fetch'):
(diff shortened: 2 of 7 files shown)


siminm:
What woudl the syntax look like for multiple options?

diranged:
@siminm, see the documentation i just added.. the way the code works, you can just keep passing in `-o option=value -o opt2=value2` (and same for the parameters).

diranged:
@siminm ok this is actually ready for review now I think

kingpin/actors/test/test_misc.py
```diff
@@ -38,6 +38,17 @@ def test_init(self):
             self.assertEquals(schema_validate.call_count, 1)
             self.assertEquals(actor.initial_actor, get_actor())
 
+    def test_init_raw_json(self):
+        json_script = ('{"desc": "",'
```
siminm: This is weird. Create a dictionary instead and then convert it to JSON

kingpin/bin/deploy.py
```diff
@@ -71,25 +75,34 @@ def main():
...
-    except Exception as e:
-        kingpin_fail(
-            '%s You must specify --json or provide it as first argument.' % e)
+    # Sanity check - did the user supply both a JSON script && an individual
+    # actor? If so, print the help!
+    if args.json and args.actor:
+        kingpin_fail('You may only specify --actor or --json, not both!')
 
+    if args.actor:
+        json_file = utils.get_script_from_args(args)
```
siminm: I'm not sure I like this approach. `arg.actor` is telling you which actor to use, but instead of using it you create a Macro that uses the actor. I recommend rewriting this part to overwrite `dry_actor` and `runner` with the actual actor that is being executed.

kingpin/actors/misc.py
```diff
@@ -166,6 +166,12 @@ def _get_macro(self):
         open the local file and return a buffer to that file.
         """
 
+        if self.option('macro').startswith('{'):
```
siminm: With my suggestions below - this part would not be needed at all.

diranged:
#310 is better ... closing this.

Issue #306, opened 2015-12-09
aws.elb.DeregisterInstance should wait for connections to drain


diranged:
I'm not sure how to do this -- but we should be able to poll the ELB and know that we're waiting for "connection draining" to happen. When that happens, we should be able to detect it and wait until its completed before returning.

PR #308, opened 2015-12-11
Issue 306: Wait for all ELB connections to drain.

diranged:
By default, the safe thing to do is to determine the "connection
draining" time configured in the ELB and wait until that time has passed
before we return from the actor. Without this, a host might move on to
another state while still handling web requests.

### DIFF
--- a/kingpin/actors/aws/elb.py
+++ b/kingpin/actors/aws/elb.py
@@ -464,7 +464,7 @@ class DeregisterInstance(base.AWSBaseActor):
     **Options**
 
     :elb:
-      (str) Name of the ELB
+      (str) Name of the ELB. Optionally this may also be a `*`.
 
     :instances:
       (str, list) Instance id, or list of ids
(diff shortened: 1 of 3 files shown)


diranged:
Also closes issue #306

kingpin/actors/aws/elb.py
```diff
@@ -506,11 +525,65 @@ def _remove(self, elb, instances):
...
+        if not self.option('wait_on_draining'):
+            self.log.warning('Not waiting for connections to drain!')
+
+        attrs = yield self.thread(elb.get_attributes)
+        if attrs.connection_draining.enabled:
+            timeout = attrs.connection_draining.timeout
+
+            self.log.info('Connection Draining Enabled, waiting %s(s)'
+                          % timeout)
+            yield utils.tornado_sleep(timeout)
```
siminm: Simple! I like it!

kingpin/actors/aws/elb.py
```diff
@@ -506,11 +525,65 @@ def _remove(self, elb, instances):
...
+            a list of LoadBalancer objects
+        """
+        all_elbs = yield self.thread(self.elb_conn.get_all_load_balancers)
+        elbs_with_members = []
+
+        for instance in instances:
+            elbs = filter(lambda lb: instance in [i.id for i in lb.instances],
+                          all_elbs)
+            self.log.debug('%s is a member of %s' % (instance, elbs))
+            elbs_with_members.extend(elbs)
```
siminm: Kinda funny that we're not leveraging the generator functionality of python here. Instead of fetching them all and returning a list we could actually yield one elb at a time.

Not worth changing things IMO because we've been doing this for a while now

siminm:
:+1: Convert the integration test to use this?

PR #316, opened 2015-12-21
Allow stringified-booleans to be passed in to the ELB actor.

diranged:


diranged:
@nrvale0 can you do a quick review here?

nrvale0:
:+1:

PR #320, opened 2016-01-12
Allow a list of channels for slack.Message

siminm:
@diranged @nrvale0 @bemsha

### DIFF
--- a/kingpin/actors/slack.py
+++ b/kingpin/actors/slack.py
@@ -33,6 +33,7 @@ sending the messages.
 
 import logging
 import os
+import re
 
 from tornado import gen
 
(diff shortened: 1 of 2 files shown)


kingpin/actors/slack.py
```diff
@@ -161,25 +162,37 @@ def _execute(self):
...
-            link_names=1,
-            unfurl_links=True,
-            unfurl_media=True
-        )
-        self._check_results(ret)
-
-        raise gen.Return()
+        # If only one channel was supplied as string then prepare the list
+        if type(self.option('channel')) == list:
+            self.log.info('channel is a list')
```
diranged: Remove debug strings

diranged:
Other than the extra log messages, :+1:

PR #321, opened 2016-01-12
Fix hipchat integration tests

siminm:


siminm:
@diranged review?

diranged:
Does it work? :)

siminm:
it does!

PR #323, opened 2016-01-13
Rewrite async test to make pypy happy

siminm:
Instead of checking sleep-time correctness I check order of execution
correctness.

Also this saves us 0.3s during unit tests :)

@diranged


diranged:
:+1:

PR #326, opened 2016-02-03
Fix rightscale concurrency execution error with single array

siminm:
cc: @diranged @bemsha @niallo


nrvale0:
:+1: :neckbeard:

kingpin/actors/rightscale/server_array.py
```diff
@@ -1326,6 +1326,9 @@ def _execute_array_with_concurrency(self, arrays, inputs):
             inputs: A string of inputs generated by
```
niallo: I'd just update the docstring as well, since it will also accept a non-list type with this change.

diranged:
:+1:merge it, then rebuild the pre_release tag. remember to delete the old pre_release zip first.

PR #328, opened 2016-02-05
add functools32 as a test dep

niallo:
Needed for sphinx build.

Please review @diranged @siminm


siminm:
I just did a fresh clone and was able to run tests without needing this library.

diranged:
I don't have a problem adding this -- its possible that certain environments have it and certain ones don't. :+1:

PR #333, opened 2016-04-18
Create rightscale.MCI actor for managing MCIs

mikhail:
cc @diranged 

Just opening this to keep track of changes. Please do not review this atrocity yet!

### DIFF
--- a/kingpin/actors/rightscale/base.py
+++ b/kingpin/actors/rightscale/base.py
@@ -42,6 +42,7 @@ import mock
 
 from kingpin.actors import base
 from kingpin.actors import exceptions
+from kingpin.actors.utils import dry
 from kingpin.actors.rightscale import api
 
 log = logging.getLogger(__name__)
@@ -171,10 +172,6 @@ class RightScaleBaseActor(base.BaseActor):
         Returns:
             A list of tuples of key/value pairs.
         """
-        if not type(params) == dict:
-            raise exceptions.InvalidOptions(
-                'Parameters passed in must be in the form of a dict.')
-
         # Nested loop that compresses a multi level dictinary into a flat
         # array of key=value strings.
         def flatten(d, parent_key=prefix, sep='_'):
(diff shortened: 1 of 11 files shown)


kingpin/actors/rightscale/base.py
```diff
@@ -195,3 +197,18 @@ def flatten(d, parent_key=prefix, sep='_'):
...
+        if not RightScaleBaseActor.account_name:
+            ca_resource = self._client._client.cloud_accounts
+            cloud_accounts = yield self._client.show(ca_resource)
+            account = yield self._client.show(cloud_accounts[0].account)
+            RightScaleBaseActor.account_name = account.soul['name']
+            log.warn('RightScale account name: %s' % account.soul['name'])
+
+    @gen.coroutine
+    def execute(self, *args, **kwargs):
+        self._log_account_name()
```
siminm: @diranged Not happy with this :( Would appreciate some feedback here.

I wanted the account name to be printed exactly once, on the first use of *any* rightscale actor. This implementation does it.

But then I would have to mock this out in **every single test** :( So what I did is remove the yield here. 

Can't have it in __init__ because it would be synchronous and blocking and I would still have to mock it in every test.

diranged: Mocking it out in every test shouldn't be too hard -- just mock it out in the `setUp()` method of the test suite?

kingpin/actors/rightscale/settings.py
```diff
@@ -44,7 +47,14 @@ def is_retriable_exception(exception):
         return False
 
     # Boto exceptions should have a code attribute
-    return str(exception) not in not_retry_codes
+    log.debug('Comparing "%s" to "%s".' % (str(exception), not_retry_codes))
+    skip_retry = any(code in str(exception) for code in not_retry_codes)
```
diranged: Doesn't this mean it would also catch `client` or `error`? That seems a bit too wide of a net to me?

diranged:
@mikhail, I think this is actually ready for review now..

kingpin/actors/rightscale/mci.py
```diff
@@ -22,137 +22,198 @@
...
+            collection=self._client._client.multi_cloud_images,
+            name=name,
+            revision=0)
+
+        # Default searches return us an empty list if there are no matching
+        # resources, or return us the exact resource we're looking for. Thus,
+        # if the return value is not a list, then we know we got the MCI back.
+        # got a list bac
+        if not isinstance(mci, list):
+            raise gen.Return(mci)
```
mikhail: I hate this part of find_by_name_and_keys

mikhail:
Looks good. :shipit:

Issue #56, opened 2014-09-04
Integration tests must be optional


mikhail:
Users that have AWS but don't have RightScale cannot currently run `make integration`

PR #335, opened 2016-04-20
Issue #56: Make each Integration test suite optional.

diranged:
This commit finally closes issue #56 -- making each set of integration
tests optional by passing different attributes to the test suite runner.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,9 @@ BUILD_DIRS = bin .build build include lib lib64 man share package *.egg
 
 .PHONY: all build clean test docs
 
+# Only execute a subset of our integration tests by default
+INTEGRATION_TESTS ?= aws,rightscale,http,rollbar,slack,pingdom
+
 all: build
 
 build: .build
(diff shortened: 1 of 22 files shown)


Makefile
```diff
@@ -24,7 +27,8 @@ test: build docs
 	python setup.py test pep8 pyflakes
 
 integration: build
-	PYFLAKES_NODOCTEST=True python setup.py integration pep8 pyflakes
+	INTEGRATION_TESTS=$(INTEGRATION_TESTS) PYFLAKES_NODOCTEST=True \
+		python setup.py integration pep8 pyflakes
```
siminm: If you run nosetest directly you can specify the integration flag plus the particular directory which you want to invoke. Much simpler than having to tag each test with a unique name

diranged: That works -- I agree -- but I don't actually think its simpler for most people. This setup allows people to run entire suites of tests with a simple command, rather than having to find each of the integration test python files.

diranged:
@siminm review?

siminm:
👍

PR #338, opened 2016-04-23
Bugfix on IAM Group deletion.

diranged:
if we're ensuring that a group is absent, its OK if we get a 404 when
searching for the groups members.

Heads up @siminm, im going to merge this so our integration tests start working again.

### DIFF
--- a/kingpin/actors/aws/iam/entities.py
+++ b/kingpin/actors/aws/iam/entities.py
@@ -742,8 +742,9 @@ class Group(EntityBaseActor):
             users = [user['user_name'] for user in
                      raw['get_group_response']['get_group_result']['users']]
         except BotoServerError as e:
-            raise exceptions.RecoverableActorFailure(
-                'An unexpected API error occurred: %s' % e)
+            if e.status != 404:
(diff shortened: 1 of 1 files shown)


kingpin/actors/aws/iam/entities.py
```diff
@@ -742,8 +742,9 @@ def _get_group_users(self, name):
             users = [user['user_name'] for user in
                      raw['get_group_response']['get_group_result']['users']]
         except BotoServerError as e:
-            raise exceptions.RecoverableActorFailure(
-                'An unexpected API error occurred: %s' % e)
+            if e.status != 404:
```
siminm: I don't think this is right. Returning empty list [] when there are no users is OK, but saying "This group has no users" when **there is no group** is wrong.

This method should raise `iam.GroupNotFound` exception and the invoking method (which *may or may not* be purging) should catch that exception and handle it gracefully

diranged: Fair -- I can refactor it to do that next week..

PR #341, opened 2016-04-28
Add option to launch new instances when instances are found stranded.

ashanbrown:
We only launch instances up to the indicated maximum replacement instances.

### DIFF
--- a/kingpin/actors/rightscale/server_array.py
+++ b/kingpin/actors/rightscale/server_array.py
@@ -946,7 +946,12 @@ class Launch(ServerArrayBaseActor):
             "Number of server to launch. Default: up to array's min count"),
         'enable': (bool, False, 'Enable autoscaling?'),
         'exact': (bool, True, (
-            'Whether to search for multiple ServerArrays and act on them.')),
+            'Whether to search for multiple ServerArrays and act on them.'),),
+        'fail_on_stranded': (bool, True, 'Fail if any instances are stranded'),
+        'replace_stranded_max': (
+            (int, str), 0,
+            "Maximum number of replacements allowed for stranded servers."
+            "Default is 0 (don't try to replace any stranded servers).")
     }
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


kingpin/actors/rightscale/server_array.py
```diff
@@ -946,7 +946,11 @@ class Launch(ServerArrayBaseActor):
             "Number of server to launch. Default: up to array's min count"),
         'enable': (bool, False, 'Enable autoscaling?'),
         'exact': (bool, True, (
-            'Whether to search for multiple ServerArrays and act on them.')),
+            'Whether to search for multiple ServerArrays and act on them.'),),
+        'replace_stranded_max': (
+            (int, str), 0,
+            "Maximum number of times to launch replacements for stranded servers."
```
siminm: This isn't accurate. If you're launching 10 machines and 2 get stranded, and your `replace_stranded_max` is set to 2 then no replacements will be tried

kingpin/actors/rightscale/server_array.py
```diff
@@ -946,7 +946,11 @@ class Launch(ServerArrayBaseActor):
...
+            "Default: don't retry.")
```
siminm: Default is 0. "don't retry" is the effect

kingpin/actors/rightscale/server_array.py
```diff
@@ -986,6 +990,8 @@ def _wait_until_healthy(self, array, sleep=60):
             min_count = int(array.soul['elasticity_params']
                             ['bounds']['min_count'])
 
+        replace_stranded_max = int(self._options.get('replace_stranded_max', 0))
```
siminm: This syntax for options is deprecated. Could you switch to `self.option('<var>')` here as well as `min_count` above?

kingpin/actors/rightscale/server_array.py
```diff
@@ -996,6 +1002,19 @@ def _wait_until_healthy(self, array, sleep=60):
             if min_count <= count:
                 raise gen.Return()
 
+            all_instances = yield self._client.get_server_array_current_instances(array)
+            stranded_instances = yield self._client.get_server_array_current_instances(
```
siminm: These two queries along with the "operational" request above constitute for 3 queries for the same info. While it might be somewhat simpler to do it this way it seems wrong to query RightScale with 3 API calls for something that can be done with just 1.

kingpin/actors/rightscale/server_array.py
```diff
@@ -996,6 +1002,19 @@ def _wait_until_healthy(self, array, sleep=60):
...
+                array, filters=['state==stranded'])
+
+            self.log.info('Found %d stranded instances.' % len(stranded_instances))
```
siminm: 99% of the time there will be 0 stranded instances, but this line doubles the log lines. I suggest putting it into an if statement and then changing this to a warning instead of info

kingpin/actors/rightscale/server_array.py
```diff
@@ -996,6 +1002,19 @@ def _wait_until_healthy(self, array, sleep=60):
...
+
+            if (len(all_instances) - len(stranded_instances) < min_count and
```
siminm: This condition is a bit hard to read. Could you provide some named vars here? If I'm reading it correctly the first conditional could just use `len(instances)` ?

siminm:
Few canonical improvements needed. Tests and RTD missing

stovoy:
It doesn't look like this does anything as to not re-trigger the extra launch over and over when the same stranded server is detected on every iteration of the boot loop. Am I not reading that right?

ashanbrown:
@stovoy My belief is that once a new instance is launched, the array should report a new server in the 'booting' state (or maybe 'queued').  I believe one of the tests verifies this.  If we discover through testing that we're spinning up too many servers, we can deal with that.  If we set the number of extra servers to just 1, it won't matter anyway.

stovoy:
Ah, ok. 👍 from me.

nextdoor-ops-ro:
Do not merge this until we have a discussion pls.

Matt Wise
Sr. Systems Architect
Nextdoor.com

PR #343, opened 2016-04-29
Add actor to destroy many rightscale arrays.

ashanbrown:
Includes an exclusion mechanism to skip over arrays.


ashanbrown:
One thing I'd like feedback on is whether we should pass an 'exact' flag to make it clear that matching is not exact by default.

siminm:
Could the "exclude" parameter be part of the exiting Terminate actor? https://github.com/Nextdoor/kingpin/blob/master/kingpin/actors/rightscale/server_array.py#L607

diranged:
So we havn't touched this in months. I vaguely remember the motivation here, but I'm not sure we ever agreed this was really necessary. Do we still think this type of an actor is useful?

Issue #342, opened 2016-04-29
Proposal -- "ensure-able" actors should treat `undef` as a sign to not manage something.


diranged:
So right now, the `aws.iam.User/Role/InstanceProfile/Group` actors take `inline_policies` and `inline_policies_purge` as options. If you don't set `inline_policies_purge` to True, then `inline_policies` just _adds_ or _modifies_ the existing policies. It does not "own" them.

After thinking about this for a few days, I think that it would be simpler and more "correct" for the actor to do the following:

If `inline_policies: undef`: Don't manage them at all. Do not delete, add or change any policies.
If `inline_policies: []`: Own the policies 100% -- and therefore, delete any policies in Amazon because there are no policies in the list.
If `inline_policies: [ "policy1", "policy2" ]`: Again, own them 100% -- add the two policies and validate that they match exactly whats in Amazon.

The advantage of this pattern is that its simpler to start off your codebase just "creating a user" like this -- and if your user already exists, then this becomes a simple and safe noop. You don't need to think about policies yet..

```
{ "actor": "aws.iam.User",
  "name": "testuser" }
```

Then when you're ready, the moment you want to manage policies you can add in one line:
```
{ "actor": "aws.iam.User",
  "name": "testuser",
  "inline_policies": [] }
```

This to me is pretty obvious that you're saying "ensure that the user has no inline policies".

@siminm, thoughts? The purpose of this issue is to not just define this for the aws actor, but more generally to define the pattern we want to follow. If you're in agreement, I'll work up the change, AND work up some documentation around this pattern.

diranged:
This is done.

PR #344, opened 2016-04-29
Issue 342 -- Remove explicit 'purge' options from AWS IAM Actors

diranged:
We've decided to refactor this code so that if a user simply does not
supply `inline_policies` or `groups` to the actors, then they
automatically do not manage these attributes. If they DO supply either
of these, then we automatically own the resource attributes entirely.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -391,6 +391,64 @@ Actors can:
         res = yield self._post_web_call(URL)
         raise gen.Return(res)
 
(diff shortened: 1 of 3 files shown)


diranged:
Closes #342

diranged:
@siminm, @nrvale0  review?

diranged:
In-person 👍  from @siminm

PR #346, opened 2016-04-30
Create kingpin.actors.utils.dry() for wrapping methods and making them dry!

mikhail:


kingpin/actors/aws/elb.py
```diff
@@ -41,6 +41,23 @@
...
 
 
+def dry(dry_message):
+    log.info('Creating _skip_on_dry decorator with "%s"' % dry_message)
+    def _skip_on_dry(function):
+        log.info('Decorating function "%s" with _skip_on_dry' % function)
+        @gen.coroutine
+        def wrapper(self, *args, **kwargs):
+            if self._dry:
+                self.log.info(dry_message)
```
siminm: pretty big issue that i've discovered here... if you do `dry_message.format(**kwargs)` it will work only when data is passed through *kwargs* and will not work if it's passed through args... meaning...

`yield self._use_certs(elb=found_elb, arn=found_arn)` will work
`yield self._use_certs(found_elb, found_arn)` will not work

sabw8217: you probably need to do some magic with `inspect.getargspec` here to make this work.

siminm: Yeah I can make it work but it'll become unreadable

diranged:
@mikhail,
  I've updated the commit here and broken it .. :). I got it working in _most_ cases, but there is a use case where it breaks. Check out the last test in kingpin.actors.test.test_utils -- the use case of calling the decorator from within a method rather than wrapping the method is what I'm tying to support.

diranged:
@mikhail  take a look at this, lemme know what you think before I move forward with it at all..

diranged:
Overall its a wash in terms of lines-of-code difference, but it forces you to write your actors in a more consistent way with multiple private methods for handling each "change" that the actor makes. I think it strongly encourages a good design pattern that makes for more readable code.

mikhail:
Yeah I'm all for it 

👍

PR #348, opened 2016-05-02
Search for the "default" RightScale MCI HREF even in the DRY run.

diranged:
Quick review @siminm ?

### DIFF
--- a/kingpin/actors/rightscale/test/test_server_array.py
+++ b/kingpin/actors/rightscale/test/test_server_array.py
@@ -277,9 +277,8 @@ class TestUpdateNextInstanceActor(testing.AsyncTestCase):
 
         # Create the actor
         self.actor = server_array.UpdateNextInstance(
-            'Patch',
-            {'array': 'unittestarray',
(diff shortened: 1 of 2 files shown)


kingpin/actors/rightscale/test/test_server_array.py
```diff
@@ -426,31 +445,6 @@ def test_execute(self):
...
-        mocked_array.self.path = '/a/b/1234'
-
-        self.actor._check_array_inputs = mock_tornado(True)
-        self.actor._find_server_arrays = mock_tornado(mocked_array)
-
-        ret = yield self.actor.execute()
-        self.assertEquals(None, ret)
-
-    @testing.gen_test
-    def test_execute_dry_with_missing_array(self):
```
siminm: Why's this removed? Testing the missing-array case sounds useful

diranged: Re-added, good catch!

diranged:
Verbal 👍

PR #349, opened 2016-05-03
Bugfix -- Ensure parse_inline_policies handles None properly.

diranged:


diranged:
@siminm  im going to merge this right after the unit tests pass because its a nasty bug .. thankfully the integration tests caught it for us.

PR #351, opened 2016-05-04
Support both YAML and JSON for the script file.

diranged:
Supporting both YAML and JSON is pretty easy. Most of this commit is
just renaming of methods so they are more generic (script instead of
json for example). Still missing a test though for the
convert_script_to_dict() method when the YAML parser is executed and it
fails.

@siminm, could you take a stab at the final missing unit test on junk YAML?

### DIFF
--- a/kingpin/utils.py
+++ b/kingpin/utils.py
@@ -185,8 +185,7 @@ def exception_logger(func):
             return func(*args, **kwargs)
         except Exception as e:
             log.debug('Exception caught in %s(%s, %s): %s' %
-                      (func, args, kwargs, e))
-            log.debug(traceback.format_exc())
+                      (func, args, kwargs, e), exc_info=1)
             raise
     return wrapper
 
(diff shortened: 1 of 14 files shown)


kingpin/utils.py
```diff
@@ -297,48 +298,62 @@ def populate_with_tokens(string, tokens, left_wrapper='%', right_wrapper='%',
...
-                                     (json_file, e))
+        raise exceptions.InvalidScript('Error reading script %s: %s' %
+                                       (script_file, e))
 
     raw = instance.read()
     parsed = populate_with_tokens(raw, tokens)
 
+    # If the file ends with .json, use demjson to read it. If it ends with
+    # .yml/.yaml, use PyYAML. If neither, error.
+    suffix = filename.split('.')[-1].strip().lower()
```
siminm: This is kinda new requirement here. there are many cases where the file isn't ending on the particular extension. I don't think we need to govern the expected file name

diranged: How should we handle the parsing then? Try both decoders, and only fail if they both fail? If we do that, we don't get JSON or YAML-specific failure messages that help the user find the error in the code.

siminm: I think a backwards-compatible solution is ok. --json and --yaml ? otherwise it'll break all the code that leverages --json

diranged: It only breaks code that uses `--json <somefile>.<some not json ending>`. Other than that, it works fine because `--json` and `--script` do the same thing.

Do we have any specific cases where we use `--json` with a file that doesnt end in `.json`?

diranged: We could make it simply look for `.yaml/yml` and if not, fall back to assuming its JSON... that might be an OK temporary solution.

siminm: Technically we have no idea because this is opensource. But I know for a fact certain deploy macros aren't named with .json since technically they aren't valid JSON until after kingpin parses them

siminm:
image

PR #352, opened 2016-05-04
Don't print out the IAM Policy Content as log.info...

diranged:
This was too verbose -- instead, print out each policy as they're parsed
, but only in debug logging. Regular logging shouldn't print out the
policy text, as that could be really large.

Quick review @siminm?


siminm:
👍

PR #358, opened 2016-05-09
Fix the Macro Integration tests after the YAML switch.

diranged:
First, make sure that on an http-downloaded script, we set the __repr__
of that in such a way that it can be parsed like a file name. Second,
restore the behavior of the convert_script_to_dict class so that we
catch LookupErrors and wrap them in an UnrecoverableActorFailure.

Review: @siminm


siminm:
remove the print. otherwise lgtm

PR #359, opened 2016-05-09
WIP: Manage S3 buckets with Kingpin

diranged:


kingpin/actors/aws/s3.py
```diff
@@ -0,0 +1,216 @@
...
+    def _execute(self):
+        """Executes an actor and yields the results when its finished.
+
+        raises: gen.Return(True)
+        """
+        yield self._ensure_bucket()
+
+        # We don't go any further -- if we're deleting the bucket, we're done
+        # here. Just return.
+        if self.option('state') == 'absent':
```
siminm: Don't get the `if` here.

kingpin/actors/aws/s3.py
```diff
@@ -0,0 +1,216 @@
...
+                raise exceptions.RecoverableActorFailure(
+                    'Cannot delete bucket: %s' % e.message)
+
+    @gen.coroutine
+    def _execute(self):
+        """Executes an actor and yields the results when its finished.
+
+        raises: gen.Return(True)
+        """
+        yield self._ensure_bucket()
```
siminm: If we're deleting a deleted bucket this actor does not produce any logs at all. Intentional?

siminm:
few questions. otherwise looks good

kingpin/actors/aws/s3.py
diranged: You'll see.. prep for the next commit.

kingpin/actors/aws/s3.py
diranged: It actually does -- see line 196?

diranged:
Going to merge this in for now -- I will have another commit shortly that adds bucket policy management.

PR #360, opened 2016-05-10
Manage S3 Bucket policies

diranged:


kingpin/actors/aws/s3.py
```diff
@@ -201,16 +235,94 @@ def _delete_bucket(self, bucket):
...
+        raise gen.Return()
+
+    @gen.coroutine
     def _execute(self):
         """Executes an actor and yields the results when its finished.
 
         raises: gen.Return(True)
         """
-        yield self._ensure_bucket()
+        bucket = yield self._ensure_bucket()
```
siminm: Kind of a weird pattern here. Half passed-variables, half class-variables. This entire _execute could be literally 2 lines:

```
yield self._ensure_bucket()
yield self._ensure_policy()
```

Just like `self.policy` is stored you can can store self.bucket which you calculate _ensure_bucket. The docstring, as well as the comments and the final raise return are pretty useless

kingpin/actors/aws/s3.py
```diff
@@ -162,10 +183,23 @@ def _create_bucket(self):
...
+        # that will be passed back. This mock object lets us simplify the rest
+        # of our code because we can mock out the results of creating a fresh
+        # empty bucket with no policies, versions, etc.
         if self._dry:
             self.log.warning('Would have created s3://%s' %
                              self.option('name'))
-            raise gen.Return()
+
+            # Generate a fake bucket and return it
+            mock_bucket = mock.MagicMock(name=self.option('name'))
```
dragonx: I would suggest, use moto to mock boto when in dry, then you don't have to manually create this mock.  You could also use a decorator that would look at the --dry flag and mock boto when true.

siminm: We don't want to mock the entire boto. We actually check if certain things exist, or if permissions are sufficient. Our "dry" run doesn't mean "offline test of syntax and compilation". It is more of a "sanity check everything that's possible without making the actual intended change"

siminm:
Fix the test, otherwise :shipit:

diranged:
Test should be fixed.. and I'll open a separate issue to move to using `moto` for mocking boto.. its a good idea.

PR #363, opened 2016-05-11
Support deep nested token passing to macros and group actors

diranged:
This patch implements a "token passing" system for the Group and Macro actors. These special actor types can pass in the "tokens" they were initialized with to each subsequent actor, allowing that actor to have the same access as the first one. This allows for a deeply nested Macro or Group actor to leverage a variable that was available at code launch time, but not necessarily passed to it.

Though this change is not backwards-incompatible -- its definitely a large change in the way we think about drawing up large and complex JSON/YAML scripts.

Closes #362.


siminm:
👍

PR #367, opened 2016-05-12
Add Bucket Logging Configuration support to the aws.s3.Bucket actor.

diranged:
Review @Siminm

### DIFF
--- a/kingpin/constants.py
+++ b/kingpin/constants.py
@@ -14,6 +14,9 @@
 
 import jsonschema
 
+from kingpin.actors import exceptions
+
+
 __author__ = 'Mikhail Simin <mikhail@nextdoor.com>'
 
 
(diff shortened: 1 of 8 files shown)


diranged:
@siminm :+1: ??

kingpin/constants.py
```diff
@@ -46,3 +48,18 @@ class STATE(StringCompareBase):
     """
 
     valid = ('present', 'absent')
+
+
+class SchemaCompareBase(object):
```
diranged: I probably should document this, and the StringCOmpareBase in the developer.rst doc file..

siminm: What do you think about removing the check for `valid` parameter and always relying on `validate` method? The STATE would simply have `validate(self, option): return option in ('present', 'state')`

siminm:
Left you a question. You can add the developer.rst docs in a separate PR. 👍

kingpin/constants.py
diranged: I refactored the code a bit .. it can be continually improved, but lets focus on that outside of this PR.

PR #369, opened 2016-05-18
Add bucket versioning management to the aws.s3.Bucket actor

diranged:
Review @Mikhail


mikhail:
looks good. What do you think about this syntax:

```python
           yield self.thread(bucket.configure_versioning, False, dry='Would disable versioning.')
```

diranged:
I don't mind it terribly -- but what bugs me is that its specific to the AWS actor. Maybe we can make a more generic method in the `BaseActor` class though that we use instead?

PR #372, opened 2016-05-23
Bugfix -- Ensure the appropriate kwargs are used for the dry decorator.

diranged:
This fix has two parts -- one, I fix the bad calls where we were passing in
*args rather than *kwargs. Second, I moved the .format() call in the
@dry decorator so that its _always_ executed, and will fail any test case
where we execute that codepath incorrectly.


diranged:
@mikhail, no need for review.. just a heads up on this one..

PR #378, opened 2016-06-21
Add ECS actor

swaggy:
@mikhail @diranged

### DIFF
--- a/kingpin/actors/aws/ecs.py
+++ b/kingpin/actors/aws/ecs.py
@@ -1,0 +1,436 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2014 Nextdoor.com, Inc
+
+"""
+:mod:`kingpin.actors.aws.ecs`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+"""
+
+import jsonschema
+import logging
+
+from tornado import gen
+
+from kingpin import utils
+from kingpin.actors import exceptions
+from kingpin.actors.aws import base
+from kingpin.actors.utils import dry
+from kingpin.constants import REQUIRED
+
+log = logging.getLogger(__name__)
+
+__author__ = 'Steve Mostovoy <smostovoy@nextdoor.com>'
+
+# http://boto3.readthedocs.io/en/latest/reference/services/ecs.html
+TASK_DEFINITION_SCHEMA = {
+    'type': 'object',
+    'required': ['family', 'containerDefinitions'],
+    'additionalProperties': False,
+    'properties': {
+        'family': {'type': 'string'},
+        'containerDefinitions': {
+            'type': 'array',
+            'items': {
+                'type': 'object',
+                'additionalProperties': False,
+                'properties': {
+                    'name': {'type': 'string'},
+                    'image': {'type': 'string'},
+                    'cpu': {'type': 'number'},
+                    'memory': {'type': 'number'},
+                    'links': {
+                        'type': 'array',
+                        'items': {'type': 'string'}
+                    },
+                    'portMappings': {
+                        'type': 'array',
+                        'items': {
+                            'type': 'object',
+                            'additionalProperties': False,
+                            'properties': {
+                                'containerPort': {'type': 'number'},
+                                'hostPort': {'type': 'number'},
+                                'protocol': {'type': 'string'}
+                            }
+                        }
+                    },
+                    'essential': {'type': 'boolean'},
+                    'entryPoint': {
+                        'type': 'array',
+                        'items': {'type': 'string'}
+                    },
+                    'command': {
+                        'type': 'array',
+                        'items': {'type': 'string'}
+                    },
+                    'environment': {
+                        'type': 'array',
+                        'items': {
+                            'type': 'object',
+                            'additionalProperties': False,
+                            'properties': {
+                                'name': {'type': 'string'},
+                                'value': {'type': 'string'}
+                            }
+                        }
+                    },
+                    'mountPoints': {
+                        'type': 'array',
+                        'items': {
+                            'type': 'object',
+                            'additionalProperties': False,
+                            'properties': {
+                                'sourceVolume': {'type': 'string'},
+                                'containerPath': {'type': 'string'},
+                                'readOnly': {'type': 'boolean'}
+                            }
+                        }
+                    },
+                    'volumesFrom': {
+                        'type': 'array',
+                        'items': {
+                            'type': 'object',
+                            'additionalProperties': False,
+                            'properties': {
+                                'sourceContainer': {'type': 'string'},
+                                'readOnly': {'type': 'boolean'}
+                            }
+                        }
+                    },
+                    'hostname': {'type': 'string'},
+                    'user': {'type': 'string'},
+                    'workingDirectory': {'type': 'string'},
+                    'disableNetworking': {'type': 'boolean'},
+                    'privileged': {'type': 'boolean'},
+                    'readonlyRootFilesystem': {'type': 'boolean'},
+                    'dnsServers': {
+                        'type': 'array',
+                        'items': {'type': 'string'}
+                    },
+                    'dnsSearchDomains': {
+                        'type': 'array',
+                        'items': {'type': 'string'}
+                    },
+                    'extraHosts': {
+                        'type': 'array',
+                        'items': {
+                            'type': 'object',
+                            'additionalProperties': False,
+                            'properties': {
+                                'hostname': {'type': 'string'},
+                                'ipAddress': {'type': 'string'}
+                            }
+                        }
+                    },
+                    'dockerSecurityOptions': {
+                        'type': 'array',
+                        'items': {'type': 'string'}
+                    },
+                    'dockerLabels': {'type': 'object'},
+                    'ulimits': {
+                        'type': 'array',
+                        'items': {
+                            'type': 'object',
+                            'required': ['name', 'softLimit', 'hardLimit'],
+                            'additionalProperties': False,
+                            'properties': {
+                                'name': {'type': 'string'},
+                                'softLimit': {'type': 'number'},
+                                'hardLimit': {'type': 'number'}
+                            }
+                        }
+                    },
+                    'logConfiguration': {'type': 'object'},
+                    'logDriver': {'type': 'string'},
+                    'options': {'type': 'object'}
+                }
+            }
+        },
+        'volumes': {
+            'type': 'array',
+            'items': {
+                'type': 'object',
+                'additionalProperties': False,
+                'properties': {
+                    'name': {'type': 'string'},
+                    'host': {
+                        'type': 'object',
+                        'additionalProperties': False,
+                        'properties': {
+                            'sourcePath': {'type': 'string'}
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+
+class ECSBaseActor(base.AWSBaseActor):
+    """Base class for ECS actors."""
+
+    def _handle_failures(self, failures):
+        """Logs ECS failures. Raises a recoverable exception if there are any.
+
+        Args:
+            failures: list of failures from ECS api.
+        """
+        if not failures:
+            return
+
+        # TODO: When can we recover? What are the failures? How do we recover?
+        for failure in failures:
+            self.log.error(failure)
+        raise exceptions.RecoverableActorFailure()
+
+
+class RunTask(ECSBaseActor):
+    """Register and run a task on ECS.
+
+    This actor will loop indefinitely until the task is complete.
+
+    **Options**
+
+    :region:
+      AWS region (or zone) name, such as us-west-2 or eu-west-1.
+
+    :cluster:
+      ECS cluster in the region to launch the task into.
+
+    :task_definition:
+      String of path to the Task Definition file template.
+      Must be a local file path.
+      Parameters to be interpolated must be of the form %VAR%.
+
+    :parameters:
+      A dictionary of key/value pairs used to fill in the parameters for the
+      Task Definition template. Default: {}.
+
+    :count:
+      How many tasks to run. Default: 1.
+
+    :wait:
+      Whether to wait for the tasks to complete. Default: True.
+
+    **Examples**
+
+    .. code-block:: yaml
+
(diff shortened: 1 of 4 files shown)


kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,111 @@
...
+log = logging.getLogger(__name__)
+
+__author__ = 'Steve Mostovoy <smostovoy@nextdoor.com>'
+
+
+# This executor is used by the tornado.concurrent.run_on_executor()
+# decorator. We would like this to be a class variable so its shared
+# across RightScale objects, but we see testing IO errors when we
+# do this.
+EXECUTOR = concurrent.futures.ThreadPoolExecutor(10)
```
niallo: `EXECUTER` doesn't seem to be used anywhere.

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class CertNotFound(exceptions.UnrecoverableActorFailure):
```
niallo: This class doesn't seem to be used.

swaggy: Oh, yes. That is correct.

swaggy:
The code is mostly feature complete.
TODO: Unit tests
TODO: Figure out failure possibilities & recoverable conditions

nextdoor-ops-ro:
Heads up -- none of your ApI calls are wrapped properly for asynchronous calling..

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,242 @@
...
+
+        Args:
+            failures: list of failures from ECS api.
+        """
+        # TODO: When can we recover? What are the failures? How do we recover?
+        if not failures:
+            return
+        for failure in failures:
+            self.log.error(failure)
+        raise exceptions.UnrecoverableActorFailure()
```
diranged: I would suggest the `RecoverableActorFailure` class. The `UnrecoverableActorFailure` class is really the "oh shit" handler. It basically lets us wrap uncaught exceptions. In this case you are catchign the exception and appropriately logging/raising it.

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,242 @@
...
+
+    """Register and run a task on ECS.
+
+    This actor will loop indefinitely until the task is complete.
+
+    **Options**
+
+    :task_definition:
+      String of path to the Task Definition file template
+      Must be a local file path
```
diranged: All three of these lines will be concatenated in the docs.. so you should use punctuation. :)

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,242 @@
...
+      AWS region (or zone) name, such as us-west-2 or eu-west-1
+
+    :cluster:
+      ECS cluster in the region to launch the task into
+
+    :count:
+      How many tasks to run
+
+    :skip_wait:
+      Whether to skip waiting for the tasks to complete
```
diranged: Document the default

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,242 @@
...
+    This actor will loop indefinitely until the task is complete.
+
+    **Options**
+
+    :task_definition:
+      String of path to the Task Definition file template
+      Must be a local file path
+      Parameters to be interpolated must be of the form %VAR%
+
+    :parameters:
```
diranged: Without reading the code, this raises a question in my head -- "Does this include automatically passed in context tokens?". See https://github.com/Nextdoor/kingpin/blob/master/kingpin/actors/base.py#L131-L133, and ping @Mikhail for some more context.

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,242 @@
...
+    :region:
+      AWS region (or zone) name, such as us-west-2 or eu-west-1
+
+    :cluster:
+      ECS cluster in the region to launch the task into
+
+    :count:
+      How many tasks to run
+
+    :skip_wait:
```
diranged: I'd re-name this to `wait_for_completion`, or even just `wait` and default it to `True`.

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,242 @@
...
+            task_definition: dict of ECS task definition parameters.
+
+        Returns:
+            tuple: ('family', 'revision').
+        """
+        if self._dry:
+            self.log.warning('Would register task')
+            return task_definition['family'], '1'
+
+        response = self.ecs_conn.register_task_definition(**task_definition)
```
diranged: I'll only comment on this once, but you do this in several places. The `self.ecs_conn` object is a synchronous object that makes blocking API calls. You will need to wrap these calls in `self.thread()` so that they are thrown into a thread and allow the Kingpin to continue doing other work in the background.

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,242 @@
...
+
+        # Parse data from the server's response.
+        task_definition = response['taskDefinition']
+        family = task_definition['family']
+        revision = task_definition['revision']
+
+        self.log.info('Task definition {0}:{1} registered'.format(
+            family, revision))
+        return family, revision
```
diranged: Any method wrapped in `gen.coroutine()` cannot `return ...` a value. Instead, you raise it. See this example:

```
@gen.coroutine
def hello_world(self):
    ret = yield self.client.get_hello_world_message()
    raise gen.Return(ret)
```

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,242 @@
...
+    def _wait_for_tasks(self, tasks):
+        """Wait for tasks to stop on ECS.
+
+        If a task stops with a non-zero exit code,
+        this will raise an exception.
+
+        Args:
+            tasks: list of task ARNs to wait for.
+        """
+        while True:
```
diranged: Check with @Mikhail, but he wrote a nifty "I'm still waiting ...." logger tool to help with these situations.

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,242 @@
...
+            containers_for_tasks = [t['containers'] for t in response['tasks']]
+            # Compute the total count of running tasks.
+            total_count = sum(len(x) for x in containers_for_tasks)
+            stopped_count = 0
+
+            for containers in containers_for_tasks:
+                for task in containers:
+                    if task['lastStatus'] == 'STOPPED':
+                        stopped_count += 1
+                    exit_code = task.get('exitCode', None)
```
diranged: again, synchronous api call... only pointing this one out because loops and async calls get interesting in tornado. You're going to want to re-factor this a bit. @Mikhail I'm sure will help out a little bit here. :)

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,242 @@
...
+                    if exit_code and exit_code != 0:
+                        self.log.error('Task {0} exited with code {1}'.format(
+                            task['taskArn'], exit_code))
+                        raise exceptions.UnrecoverableActorFailure()
```
diranged: Recoverable... :)

kingpin/actors/aws/ecs.py
swaggy: I saw that (utils.create_repeating_log), but didn't use it because I'd like each message to say the current progress.

swaggy:
@diranged @mikhail Another review please - 100% unit test coverage for ECS RunTask actor.

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,410 @@
...
+            tuple: ('family', 'revision').
+        """
+        # Validate the schema.
+        try:
+            jsonschema.validate(task_definition, TASK_DEFINITION_SCHEMA)
+        except jsonschema.exceptions.ValidationError as e:
+            raise exceptions.InvalidOptions(e)
+
+        if self._dry:
+            self.log.warning('Would register task')
```
mikhail: Should be Info not Warning. Any information you could print here? maybe family and version?

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,410 @@
...
+            raise gen.Return((task_definition['family'], 1))
+
+        response = yield self.thread(
```
mikhail: Don't know how long this runs, but in non-dry mode there is no log before this execution.

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,410 @@
...
+        task_definition = response['taskDefinition']
+        family = task_definition['family']
+        revision = task_definition['revision']
+
+        self.log.info('Task definition {0}:{1} registered'.format(
+            family, revision))
+        raise gen.Return((family, revision))
+
+    @gen.coroutine
+    @dry('Would run task {0}:{1}')
```
mikhail: Doubt this works. Instead use the variable names: Would run task {family}:{revision}

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,410 @@
...
+    def _wait_for_tasks(self, tasks):
+        """Wait for tasks to stop on ECS.
+
+        If a task stops with a non-zero exit code,
+        this will raise an exception.
+
+        Args:
+            tasks: list of task ARNs to wait for.
+        """
+        while True:
```
mikhail: What about this layout?

```python
done = yield self.tasks()
while not done:
    sleep 5
    done = yield self.tasks()
```

?

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,410 @@
...
+        stopped_count = 0
+
+        for container in containers:
+            if container['lastStatus'] == 'STOPPED':
+                stopped_count += 1
+            exit_code = container.get('exitCode', None)
+            if exit_code and exit_code != 0:
+                self.log.error('Task {0} exited with code {1}'.format(
+                    container['taskArn'], exit_code))
+                raise exceptions.UnrecoverableActorFailure()
```
mikhail: I think it's ok to have this be recoverable. The user of this actor should be able to decide if it's essential, and optionally leverage the warn_on_failure

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,410 @@
...
+
+        if stopped_count == total_count:
+            self.log.info('All {0} tasks stopped'.format(total_count))
+            raise gen.Return(True)
```
mikhail: Add a new line after this `if` block. These two steps aren't connected logically

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,410 @@
...
+        self.log.info('Running task from {task_definition} into ECS. '
+                      'Region: {region}, cluster: {cluster}.')
+        task_definition = utils.convert_script_to_dict(
+            self.option('task_definition'), self.option('parameters'))
+        family, revision = yield self._register_task(task_definition)
+        tasks = yield self._run_task(family, revision)
+        if self.option('wait'):
+            yield self._wait_for_tasks(tasks)
+        else:
+            self.log.info('Not waiting for tasks to stop')
```
mikhail: Should be a warning, not an info

kingpin/actors/aws/ecs.py
swaggy: Nah, I don't like repeating the call.

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,418 @@
...
+
+        Args:
+            task_definition: dict of ECS task definition parameters.
+
+        Returns:
+            tuple: ('family', 'revision').
+        """
+        # Validate the schema.
+        try:
+            jsonschema.validate(task_definition, TASK_DEFINITION_SCHEMA)
```
diranged: Do this in `self.__init__()`... that way it happens long before the execution of the code. It will fail immediately at compile time.

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,418 @@
...
+        except jsonschema.exceptions.ValidationError as e:
+            raise exceptions.InvalidOptions(e)
+
+        family = task_definition['family']
+        if self._dry:
```
diranged: Use the `@dry` wrapper instead here, then you don't need any of this if/else code..

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,418 @@
...
+            cluster=self.option('cluster'),
+            taskDefinition='{0}:{1}'.format(family, revision),
+            count=self.option('count'))
+        self._handle_failures(response['failures'])
+        tasks = [t['taskArn'] for t in response['tasks']]
+        raise gen.Return(tasks)
+
+    @gen.coroutine
+    @dry('Would wait for tasks')
+    def _wait_for_tasks(self, tasks):
```
diranged: Overall I don't love this method.. a few nits. First, it doesn't provide any feedback to the user that it "waited" and is going to check again in 5s. This is where the `repeating_log` logic that Mikhail wrote is pretty useful. Second, you wrap it in a @dry decorator when its not really necessary. If it gets an empty list of tasks (or None), it should just return.

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,418 @@
...
+        self.log.info(
+            '{0} tasks stopped out of {1} tasks'.format(
+                stopped_count, total_count))
+        raise gen.Return(False)
+
+    @gen.coroutine
+    def _execute(self):
+        self.log.info('Running task from {task_definition} into ECS. '
+                      'Region: {region}, cluster: {cluster}.')
+        task_definition = utils.convert_script_to_dict(
```
diranged: do this in `__init__`...

kingpin/actors/aws/ecs.py
```diff
@@ -0,0 +1,418 @@
...
+            tasks: list of task ARNs to wait for.
+        """
+        while True:
+            done = yield self._tasks_done(tasks)
+            if done:
+                break
+            yield gen.sleep(5)
+
+    @gen.coroutine
+    def _tasks_done(self, tasks):
```
diranged: So I'd like to see this rewritten as `self._task_done(self, task)` and then a wrapper `self._tasks_done(self, tasks)` that leverages `self._task_done()`. I know thats _kind_ of what `wait_for_tasks()` does, but really wait for tasks is a wrapper thats super simple. 

Maybe we can chat about this part in person..

kingpin/actors/aws/ecs.py
swaggy: I can't, I need it to return (family, 1)

kingpin/actors/aws/ecs.py
swaggy: Can't use repeating log - see my previous comment. It doesn't work with dynamic text.

kingpin/actors/aws/ecs.py
swaggy: Can't split it like that, or else it'd need multiple API calls. This is a single batch call.

mikhail:
This looks good! merge it!

diranged:
I'm 99% sure that there are no doc links with this commit, meaning its an undocumented actor..

PR #379, opened 2016-06-29
Update deploy.py

mikhail:


mikhail:
Review? @swaggy

kingpin/bin/deploy.py
```diff
@@ -120,7 +120,7 @@ def main():
     if args.build_only:
         try:
             get_main_actor(dry=False)
-        except actor_exceptions.ActorException as e:
+        except Exception as e:
```
diranged: Why this change?

mikhail: Because if something raises an exception that wasn't caught and converted to a kingpin exception then this code doesn't catch it and exits with status 0.

For example - yaml parser raised ScannerException which was not caught by anything

swaggy:
👍 Nice 👍

kingpin/bin/deploy.py
diranged: That was an intentional decision by you and I back in the day.. the idea was that any uncaught exception _DID_ crash the app on purpose because its not expected.

diranged: I should add, this can be rethought ... but it was intentional at the time.

mikhail: It still crashes the app ... I explicitly log it and exit with status 1. The problem is that if python exists on its own - then it's status 0, and that's bad.

diranged:
👍 lets just merge this.. seems fine i think.

PR #385, opened 2016-07-13
Add ecs service actor.

swaggy:
Add ECS service actor. Still working on unit tests, then integration tests.

@mikhail @diranged

### DIFF
--- a/kingpin/actors/aws/ecs.py
+++ b/kingpin/actors/aws/ecs.py
@@ -19,7 +19,10 @@
 
 import jsonschema
 import logging
+import operator
+import uuid
 
+from boto3.exceptions import Boto3Error
 from tornado import gen
 
 from kingpin import utils
@@ -36,14 +39,12 @@ __author__ = 'Steve Mostovoy <smostovoy@nextdoor.com>'
 TASK_DEFINITION_SCHEMA = {
     'type': 'object',
     'required': ['family', 'containerDefinitions'],
-    'additionalProperties': False,
     'properties': {
         'family': {'type': 'string'},
         'containerDefinitions': {
             'type': 'array',
             'items': {
                 'type': 'object',
-                'additionalProperties': False,
                 'properties': {
                     'name': {'type': 'string'},
                     'image': {'type': 'string'},
@@ -57,7 +58,6 @@ TASK_DEFINITION_SCHEMA = {
                         'type': 'array',
                         'items': {
                             'type': 'object',
-                            'additionalProperties': False,
                             'properties': {
                                 'containerPort': {'type': 'number'},
                                 'hostPort': {'type': 'number'},
@@ -78,7 +78,6 @@ TASK_DEFINITION_SCHEMA = {
                         'type': 'array',
                         'items': {
                             'type': 'object',
-                            'additionalProperties': False,
                             'properties': {
                                 'name': {'type': 'string'},
                                 'value': {'type': 'string'}
@@ -89,7 +88,6 @@ TASK_DEFINITION_SCHEMA = {
                         'type': 'array',
                         'items': {
                             'type': 'object',
-                            'additionalProperties': False,
                             'properties': {
                                 'sourceVolume': {'type': 'string'},
                                 'containerPath': {'type': 'string'},
@@ -101,7 +99,6 @@ TASK_DEFINITION_SCHEMA = {
                         'type': 'array',
                         'items': {
                             'type': 'object',
-                            'additionalProperties': False,
                             'properties': {
                                 'sourceContainer': {'type': 'string'},
                                 'readOnly': {'type': 'boolean'}
@@ -126,7 +123,6 @@ TASK_DEFINITION_SCHEMA = {
                         'type': 'array',
                         'items': {
                             'type': 'object',
-                            'additionalProperties': False,
                             'properties': {
                                 'hostname': {'type': 'string'},
                                 'ipAddress': {'type': 'string'}
@@ -143,7 +139,6 @@ TASK_DEFINITION_SCHEMA = {
                         'items': {
                             'type': 'object',
                             'required': ['name', 'softLimit', 'hardLimit'],
-                            'additionalProperties': False,
                             'properties': {
                                 'name': {'type': 'string'},
                                 'softLimit': {'type': 'number'},
@@ -161,12 +156,10 @@ TASK_DEFINITION_SCHEMA = {
             'type': 'array',
             'items': {
                 'type': 'object',
-                'additionalProperties': False,
                 'properties': {
                     'name': {'type': 'string'},
                     'host': {
                         'type': 'object',
-                        'additionalProperties': False,
                         'properties': {
                             'sourcePath': {'type': 'string'}
                         }
@@ -181,19 +174,113 @@ TASK_DEFINITION_SCHEMA = {
 class ECSBaseActor(base.AWSBaseActor):
     """Base class for ECS actors."""
 
-    def _handle_failures(self, failures):
+    FAILURE_MISSING = 'MISSING'
+
+    def _handle_failures(self, failures, *ignorable):
         """Logs ECS failures. Raises a recoverable exception if there are any.
 
         Args:
             failures: list of failures from ECS api.
+            ignorable: list of failures to ignore.
         """
         if not failures:
             return
 
-        # TODO: When can we recover? What are the failures? How do we recover?
+        failure_reasons = []
+
         for failure in failures:
+            reason = failure['reason']
+            if reason in ignorable:
+                continue
+            failure_reasons.append(reason)
             self.log.error(failure)
-        raise exceptions.RecoverableActorFailure()
+        if failure_reasons:
+            raise exceptions.RecoverableActorFailure(failure_reasons)
+
+    @gen.coroutine
+    @dry('Would register task definition with family {0[family]}')
+    def _register_task(self, task_definition):
+        """Registers a task on ECS.
+
+        Args:
+            task_definition: dict of ECS task definition parameters.
+
+        Returns:
+            tuple: ('family', 'revision', 'task_definition_name').
+        """
+
+        family = task_definition['family']
+
+        self.log.info('Registering task definition with family {}'.format(
+            family))
+        try:
+            response = yield self.thread(
+                self.ecs_conn.register_task_definition, **task_definition)
+        except Boto3Error as e:
+            raise exceptions.RecoverableActorFailure(e.message)
+
+        # Parse data from the server's response.
+        task_definition = response['taskDefinition']
+        family = task_definition['family']
+        revision = task_definition['revision']
+
+        task_definition_name = '{}:{}'.format(family, revision)
+        self.log.info('Task definition {} registered'.format(
+            task_definition_name))
+        raise gen.Return((family, revision, task_definition_name))
+
+    @staticmethod
+    def _load_task_definition(task_definition_file, tokens):
+        """Loads and verifies a task definition template file, and interpolates
+        tokens.
+
+        Args:
+            task_definition_file: task definition file to load.
+            tokens: dict of key/value pairs to interpolate into the file.
+
+        Returns:
+            Resulting task definition dict.
+        """
+        task_definition = utils.convert_script_to_dict(
+            task_definition_file, tokens)
+        try:
+            jsonschema.validate(task_definition,
+                                TASK_DEFINITION_SCHEMA)
+        except jsonschema.exceptions.ValidationError as e:
+            raise exceptions.InvalidOptions(e)
+        return task_definition
+
+    @staticmethod
+    def _load_service_definition(service_definition_file, tokens):
+        """Loads and verifies a service definition template file, and interpolates
+        tokens. The service definition template file can be None.
+
+        Args:
+            service_definition_file: service definition file to load.
+                If None or an empty string, this returns only defaults.
+            tokens: dict of key/value pairs to interpolate into the file.
+
+        Returns:
+            Resulting service definition dict.
+        """
+        if not service_definition_file:
+            service_definition = {}
+        else:
+            service_definition = utils.convert_script_to_dict(
+                service_definition_file, tokens)
+            try:
+                jsonschema.validate(service_definition,
+                                    SERVICE_DEFINITION_SCHEMA)
+
+            except jsonschema.exceptions.ValidationError as e:
+                raise exceptions.InvalidOptions(e)
+
+        # Set default values.
+        service_definition.setdefault(
+            'loadBalancers', [])
+        service_definition.setdefault(
+            'deploymentConfiguration', {})
+        return service_definition
 
 
 class RunTask(ECSBaseActor):
@@ -212,10 +299,10 @@ class RunTask(ECSBaseActor):
     :task_definition:
       String of path to the Task Definition file template.
       Must be a local file path.
-      Parameters to be interpolated must be of the form %VAR%.
+      Tokens to be interpolated must be of the form %VAR%.
 
-    :parameters:
-      A dictionary of key/value pairs used to fill in the parameters for the
+    :tokens:
+      A dictionary of key/value pairs used to fill in the tokens for the
       Task Definition template. Default: {}.
 
     :count:
@@ -228,10 +315,10 @@ class RunTask(ECSBaseActor):
 
     .. code-block:: yaml
 
-       actor: aws.ecs.RunTask
+       actor: aws.ecs.Task
        desc: Run migrations
        options:
-          task_definition: nextdoor-migrate.json
+          task_definition: migrate.yaml
           region: us-west-2
           cluster: us1-internal
 
@@ -249,10 +336,11 @@ class RunTask(ECSBaseActor):
             (str, REQUIRED,
              'String of path to the Task Definition file template. '
              'Must be a local file path. '
-             'Parameters to be interpolated must be of the form %VAR%.'),
-        'parameters': (dict, {},
-                       'A dictionary of key/value pairs '
-                       'used to fill in the parameters for the template.'),
+             'Tokens to be interpolated must be of the form %VAR%.'),
+        'tokens': (dict, {},
+                   'A dictionary of key/value pairs '
+                   'used to fill in the tokens for the '
+                   'Task Definition template.'),
         'count': (int, 1, 'How many tasks to run.'),
         'wait': (bool, True,
                  'Whether to wait for the tasks to complete.')
@@ -260,69 +348,51 @@ class RunTask(ECSBaseActor):
 
     def __init__(self, *args, **kwargs):
         super(RunTask, self).__init__(*args, **kwargs)
-        self.task_definition = _load_task_definition(
+        self.task_definition = self._load_task_definition(
             self.option('task_definition'),
-            self.option('parameters'))
+            self.option('tokens'))
 
     @gen.coroutine
-    def _register_task(self, task_definition):
-        """Registers a task on ECS.
-
-        Args:
-            task_definition: dict of ECS task definition parameters.
-
-        Returns:
-            tuple: ('family', 'revision').
-        """
-        family = task_definition['family']
-        if self._dry:
-            self.log.info(
-                'Would register task definition with family {0}'.format(
-                    family))
-            raise gen.Return((family, 1))
-
-        self.log.info('Registering task definition with family {0}'.format(
-            family))
-        response = yield self.thread(
-            self.ecs_conn.register_task_definition, **task_definition)
-
-        # Parse data from the server's response.
-        task_definition = response['taskDefinition']
-        family = task_definition['family']
-        revision = task_definition['revision']
-
-        self.log.info('Task definition {0}:{1} registered'.format(
-            family, revision))
-        raise gen.Return((family, revision))
-
-    @gen.coroutine
-    @dry('Would run task {family}:{revision}')
-    def _run_task(self, family, revision):
+    @dry('Would run task {0}')
+    def _run_task(self, task_definition_name):
         """Runs a task on ECS given a task definition's family and revision.
 
         A task can result in multiple running tasks,
         depending on count and sidekick tasks.
 
         Args:
-            family: Task family string
-            revision: Task revision string
+            task_definition_name: Task Definition string
 
         Returns:
             list: task ARNs.
         """
-        response = yield self.thread(
-            self.ecs_conn.run_task,
-            cluster=self.option('cluster'),
-            taskDefinition='{0}:{1}'.format(family, revision),
-            count=self.option('count'))
-        self.log.info('Scheduled task {0}:{1}'.format(family, revision))
-        self._handle_failures(response['failures'])
+        repeating_log = utils.create_repeating_log(
+            self.log.info,
+            'Waiting for task to be found...',
+            seconds=30)
+
+        while True:
+            response = yield self.thread(
+                self.ecs_conn.run_task,
+                cluster=self.option('cluster'),
+                taskDefinition=task_definition_name,
+                count=self.option('count'))
+
+            if not response['failures']:
+                break
+            # Error on non-missing failures.
+            self._handle_failures(response['failures'], self.FAILURE_MISSING)
+            yield gen.sleep(2)
+
+        utils.clear_repeating_log(repeating_log)
+
+        self.log.info('Scheduled task {}'.format(task_definition_name))
         tasks = [t['taskArn'] for t in response['tasks']]
         raise gen.Return(tasks)
 
     @gen.coroutine
     def _wait_for_tasks(self, tasks):
-        """Wait for tasks to stop on ECS.
+        """Wait for tasks to complete on ECS.
 
         If a task stops with a non-zero exit code,
         this will raise an exception.
@@ -336,7 +406,7 @@ class RunTask(ECSBaseActor):
             done = yield self._tasks_done(tasks)
             if done:
                 break
-            yield gen.sleep(5)
+            yield gen.sleep(10)
 
     @gen.coroutine
     def _tasks_done(self, tasks):
@@ -355,7 +425,7 @@ class RunTask(ECSBaseActor):
 
         self._handle_failures(response['failures'])
 
-        containers = _get_containers_from_tasks(
+        containers = self._get_containers_from_tasks(
             task_list=response['tasks'])
 
         total_count = len(containers)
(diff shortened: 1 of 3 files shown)


kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,399 @@ def _execute(self):
...
+             "Used for: 'loadBalancers', 'role', 'deploymentConfiguration'"),
+        'service_name':
+            (str, None,
+             'Service name to use. If not specified, this will use '
+             "the Task Definition's family."),
+        'tokens': (dict, {},
+                   'A dictionary of key/value pairs '
+                   'used to fill in the tokens for the Task and Service '
+                   'definition templates.'),
+        'count': (int, 1, 'How many services to run.'),
```
mikhail: Tasks or services? This comment contradicts the docstring above

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,399 @@ def _execute(self):
...
+        response = yield self.thread(
+            self.ecs_conn.describe_services,
+            cluster=self.option('cluster'),
+            services=[service_name])
+        self._handle_failures(response['failures'], self.FAILURE_MISSING)
+
+        services = response['services']
+
+        # There should never be more than one service for a given name.
+        assert len(services) < 2
```
mikhail: `AssertionError` will break the entire python code here. and it will not provide any details.

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,399 @@ def _execute(self):
...
+        """
+        create_parameters = dict(
+            cluster=self.option('cluster'),
+            serviceName=service_name,
+            taskDefinition='{0}:{1}'.format(family, revision),
+            desiredCount=self.option('count'),
+            clientToken=client_token,
+            **self.service_definition)
+
+        if self._dry:
```
mikhail: Might as well use a decorator for the entire function...

mikhail:
This is starting to look really good

mikhail:
This looks good to me, but, @diranged, can you take a look also?

nextdoor-ops-ro:
Will review tomorrow morning

swaggy:
@diranged Have a chance to review?

kingpin/actors/aws/ecs.py
```diff
@@ -181,19 +183,60 @@
...
 
-    def _handle_failures(self, failures):
+    FAILURE_MISSING = 'MISSING'
+
+    def _handle_failures(self, failures, *ignoreable):
         """Logs ECS failures. Raises a recoverable exception if there are any.
 
         Args:
             failures: list of failures from ECS api.
+            ignoreable: list of failures to ignore.
```
niallo: s/ignoreable/ignorable/g

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+            services=[service_name])
+        self._handle_failures(response['failures'], self.FAILURE_MISSING)
+
+        services = response['services']
+
+        # There should never be more than one service for a given name.
+        if len(services) > 1:
+            self.log.warning(
+                'API returned more than one service for {name}'.format(
+                    name=service_name))
```
niallo: Should you also warn (or error) on `len(services) == 0`?

swaggy: See line 665 for usage, that's an expected case.
I updated the DocString of describe_services to be more explicit here.

kingpin/actors/aws/ecs.py
```diff
@@ -181,19 +183,60 @@
...
+
+        Args:
+            task_definition: dict of ECS task definition parameters.
+
+        Returns:
+            tuple: ('family', 'revision').
+        """
+
+        family = task_definition['family']
+        if self._dry:
```
diranged: Instead of 5 lines here, use the `@dry` decorator, and make your calling code smart enough to handle `None` as a return value.

kingpin/actors/aws/ecs.py
```diff
@@ -181,19 +183,60 @@
...
+            self.log.info(
+                'Would register task definition with family {0}'.format(
+                    family))
+            raise gen.Return((family, 1))
+
+        self.log.info('Registering task definition with family {0}'.format(
+            family))
+        response = yield self.thread(
```
diranged: What happens if this fails -- the inputs are bad, a 500 is raised, etc? I don't see any catching here or in the calling-code.

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
             yield self._wait_for_tasks(tasks)
         else:
-            self.log.info('Not waiting for tasks to stop')
+            self.log.info('Not waiting for tasks to complete')
+
+
+# http://boto3.readthedocs.io/en/latest/reference/services/ecs.html
+SERVICE_DEFINITION_SCHEMA = {
+    'type': 'object',
+    'additionalProperties': False,
```
diranged: I would re-think `additionalProperties: False` -- with this in place, every time Amazon adds some new feature to ECS (Which they're doing regularly), we are forced to update our code. I would switch this to `True` here and in the TASK definition scema.

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+            'properties': {
+                'maximumPercent': {'type': 'number'},
+                'minimumHealthyPercent': {'type': 'number'},
+            }
+        }
+    }
+}
+
+
+class DeployService(ECSBaseActor):
```
diranged: I would rename this to `Service`. You're really ensuring that a given service is running with a given task definition. If you were simply "deploying" the service (where a task definition had already been defined), then I can see this being named `DeployService`.

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+      Must be a local file path.
+      Tokens to be interpolated must be of the form %VAR%.
+
+    :service_definition:
+      String of path to the Service Definition file template.
+      Must be a local file path.
+      Tokens to be interpolated must be of the form %VAR%.
+      Implicit fields - do not include these:
+      'serviceName', 'taskDefinition', 'desiredCount', 'clientToken'
+      Used for: 'loadBalancers', 'role', 'deploymentConfiguration'
```
diranged: This verbage is confusing -- I'm not sure what you're saying here?

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+      A dictionary of key/value pairs used to fill in the tokens for the
+      Task and Service Definition template. Default: {}.
+
+    :count:
+      How many instances of the service to deploy.
+      Default: 1.
+
+    :wait:
+      Whether to wait for the services to deploy.
+      Default: True.
```
diranged: I would make this a time (in seconds)... so that each person can define how long they want to wait before erroring out.

kingpin/actors/aws/ecs.py
diranged: Shouldn't we be raising an exception here?

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+            cluster=self.option('cluster'),
+            serviceName=service_name,
+            taskDefinition='{0}:{1}'.format(family, revision),
+            desiredCount=self.option('count'),
+            clientToken=client_token,
+            **self.service_definition)
+
+        self.log.info('Creating service.')
+
+        yield self.thread(
```
diranged: Exception catching ... what happens if there is a failure here? Can we tell the user a bit about the failure?

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+        update_parameters = dict(
+            cluster=self.option('cluster'),
+            service=service_name,
+            taskDefinition='{0}:{1}'.format(family, revision),
+            desiredCount=self.option('count'),
+            deploymentConfiguration=deployment_configuration)
+
+        self.log.info('Updating service.')
+
+        yield self.thread(
```
diranged: More exception catching here?

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+            self.ecs_conn.update_service,
+            **update_parameters)
+
+    @gen.coroutine
+    def _register_service(self, service_name, family, revision):
```
diranged: Would you mind renaming this to `_ensure_service()`? That follows the naming I've been using in the other actors that "ensure a resource is in a given state".

kingpin/actors/aws/ecs.py
```diff
@@ -422,31 +798,126 @@ def _get_containers_from_tasks(task_list):
...
-        jsonschema.validate(task_definition_file,
+        jsonschema.validate(task_definition,
                             TASK_DEFINITION_SCHEMA)
     except jsonschema.exceptions.ValidationError as e:
         raise exceptions.InvalidOptions(e)
-    return task_definition_file
+    return task_definition
+
+
+def _load_service_definition(service_definition_file, tokens):
```
diranged: I'm not a fan of these utility functions not being inside one of the classes. Is there a good reason they aren't either a) inside a utility package, or b) inside the ECSBaseActor class?

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+        else:
+            errors = _get_immutable_field_errors(
+                old_params=existing_service,
+                new_params=self.service_definition,
+                immutable_fields=['loadBalancers', 'role'])
+
+            if errors:
+                for error in errors:
+                    self.log.error(error)
+                raise exceptions.UnrecoverableActorFailure(
```
diranged: This is still a `RecoverableActorFailure` -- its an expected and planned for exception. The user can choose to ignore this and let the rest of their Kingpin script continue if they wish.

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+            yield self._create_service(service_name,
+                                       family, revision,
+                                       client_token)
+        else:
+            errors = _get_immutable_field_errors(
+                old_params=existing_service,
+                new_params=self.service_definition,
+                immutable_fields=['loadBalancers', 'role'])
+
+            if errors:
```
diranged: Really, I'd just move the failure/exception logic into `_get_immutable_field_errors()` -- then it turns into `_verify_no_immutable_field_changes()` and its a one-liner here in this function.

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+        while True:
+            done = yield self._service_deployed(
+                service_name,
+                seen_events)
+            if done:
+                break
+            yield gen.sleep(10)
+
+    @gen.coroutine
+    def _service_deployed(self, service_name, seen_events):
```
diranged: Not a fan of this function name -- it sounds like its saying "the service IS deployed", not "checking IF the service is deployed". Sure, once I read it, I get it.. but its not obvious at first glance.

Rename it to `self._is_service_deployed()`?

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+
+        deployments = service['deployments']
+        primary_deployment = None
+        for deployment in deployments:
+            if deployment['status'] == 'PRIMARY':
+                primary_deployment = deployment
+                break
+        if not primary_deployment:
+            # There should always be one 'PRIMARY' deployment returned.
+            raise exceptions.UnrecoverableActorFailure(
```
diranged: Recoverable..

kingpin/actors/aws/ecs.py
```diff
@@ -422,31 +798,126 @@ def _get_containers_from_tasks(task_list):
...
+                'Old value: {old}\n'
+                'New value: {new}'.format(
+                    field=immutable_field_name,
+                    old=old_field,
+                    new=new_field))
+
+    return errors
+
+
+def _get_sorted_new_log_events(events, start_timestamp, seen_events):
```
diranged: I just LOVE that Amazon makes us do this...

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+
+        Args:
+            service_name: name of the service to wait for.
+        """
+        # Used to ensure event logs are only printed once.
+        seen_events = set()
+        while True:
+            done = yield self._service_deployed(
+                service_name,
+                seen_events)
```
diranged: So I sort of hate this, but I can't think of an obvious way to fix it right now.. but I don't like that you're passing this single `seen_events` set around through three different functions, but without reading all three functions you can't really grasp that its the _same_ set being passed around and modified. This gives me the same heebie-jeebies as nested functions -- they require tons of documentation to make it really really clear whats going on.

Lets take a few minutes today and talk through this and see if we can come up with a more elegant solution here.

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+            if done:
+                break
+            yield gen.sleep(10)
+
+    @gen.coroutine
+    def _service_deployed(self, service_name, seen_events):
+        """Checks if service is finished deploying.
```
diranged: I think that this function needs more documentation that really indicates that its meant to be used in a loop from an outside caller.

kingpin/actors/aws/ecs.py
```diff
@@ -406,7 +420,369 @@ def _execute(self):
...
+            self.log.info('Event: {message}'.format(
+                message=event_message))
+
+        running_count = primary_deployment['runningCount']
+        desired_count = primary_deployment['desiredCount']
+        missing_count = desired_count - running_count
+
+        extra_deployment_count = len(deployments) - 1
+
+        if missing_count == 0 and extra_deployment_count == 0:
```
diranged: So I'm not sure if this is a realistic problem -- but by only checking that the number of deployments is correct, I think that we could see a race condition where the Amazon API returns stale data. I could see a situation where we call this method _right_ after we call the `_update_service()` method, and they return only the "primary" deployment and not the "new" deployment thats being set up.

Can we add some more checking here to make sure that the "primary deployment" task ARN matches what we expect?

diranged:
Hey @swaggy, functionally looks great.. I have some style nits in there that I'd like to fix up, and one or two areas that I'd like to see improved. Finally, you're still missing doc string references I believe, meaning that the Sphinx docs will not build links for the ECS actors in our main docs.. need to fix that up too.

swaggy:
@diranged thanks for the review Matt, will address your comments soon.

kingpin/actors/aws/ecs.py
swaggy: Can't, we also want to log task_definition['family'] in there

kingpin/actors/aws/ecs.py
swaggy: The function documentation doesn't explicitly specify any exceptions / errors that can occur. I've wrapped it in a try catch for a boto3 error.

kingpin/actors/aws/ecs.py
diranged: Sure you can -- you can do that inside the `@dry` decorator like this:

```
@dry('Would register task definition with family {task_definition[family]}.')
```

kingpin/actors/aws/ecs.py
swaggy: I disagree, but only because the schema is linked to the version of Boto we are using. We will have to maintain this along with boto3 but I think it's worth it.

kingpin/actors/aws/ecs.py
swaggy: 👍

kingpin/actors/aws/ecs.py
swaggy: 'serviceName', 'taskDefinition', 'desiredCount', 'clientToken' are figured out by the actor and should not be included in the service definition. Changed 'used for' to 'allowed fields'

kingpin/actors/aws/ecs.py
swaggy: Same as above, I can wrap it in a boto3 error handler

kingpin/actors/aws/ecs.py
swaggy: Ditto

kingpin/actors/aws/ecs.py
mikhail: We have a global actor timeout value. I'd keep this as boolean and encourage using the `timeout` param : http://kingpin.readthedocs.io/en/stable/basicuse.html?highlight=timeout#schema-description

kingpin/actors/aws/ecs.py
swaggy: Yeah it's great. 💩

kingpin/actors/aws/ecs.py
diranged: Fair enough, thats fine..

kingpin/actors/aws/ecs.py
swaggy: Moved the event logging to the actor and made seen_events a field in the actor (self.seen_events).

kingpin/actors/aws/ecs.py
swaggy: Good call. Let's just throw an exception in that case for now and see if it comes up.

kingpin/actors/aws/ecs.py
diranged: Yeah so this is the situation where I would typically start sending bogus data to them and trying to fail the service intentionally...

kingpin/actors/aws/ecs.py
diranged: Are you sure? I was pretty sure that they just pass the JSOn right on up to the API and let the API deal with it.

swaggy:
Fixing the docs too.

kingpin/actors/aws/ecs.py
swaggy: Oh, cool.

kingpin/actors/aws/ecs.py
swaggy: You're right. I've removed additional properties.

kingpin/actors/aws/ecs.py
mikhail: This doesn't behave like the Boneman actors do. I think the name is fine as is. This actor will deploy a new service every time it's run, even if all the parameters are same as previous run. It's not a state keeper like S3 or IAM actors

swaggy:
@diranged This is now running live on staging and everything seems to be working.

kingpin/actors/aws/ecs.py
```diff
@@ -181,19 +174,113 @@
...
+
+        family = task_definition['family']
+
+        self.log.info('Registering task definition with family {}'.format(
+            family))
+        try:
+            response = yield self.thread(
+                self.ecs_conn.register_task_definition, **task_definition)
+        except Boto3Error as e:
+            raise exceptions.UnrecoverableActorFailure(e.message)
```
diranged: Lets make this `Recoverable`.. admittedly we're not sure what error we're catching, but at least we're catching it.

kingpin/actors/aws/ecs.py
```diff
@@ -181,19 +174,113 @@
...
+        task_definition = response['taskDefinition']
+        family = task_definition['family']
+        revision = task_definition['revision']
+
+        task_definition_name = '{}:{}'.format(family, revision)
+        self.log.info('Task definition {} registered'.format(
+            task_definition_name))
+        raise gen.Return((family, revision, task_definition_name))
+
+    @staticmethod
```
diranged: Ohh... fancy pants. I think this is the first one of these in the entire code base!

diranged: I won't make you change it -- but do me a favor, explain _why_ this is a static method? I know functionally that it _can be_ a static method, but I'm curious what the advantage is?

kingpin/actors/aws/ecs.py
```diff
@@ -181,19 +174,113 @@
...
+
+        Args:
+            task_definition_file: task definition file to load.
+            tokens: dict of key/value pairs to interpolate into the file.
+
+        Returns:
+            Resulting task definition dict.
+        """
+        task_definition = utils.convert_script_to_dict(
+            task_definition_file, tokens)
```
diranged: I would recommend leveraging the `init_tokens` (https://github.com/Nextdoor/kingpin/search?utf8=%E2%9C%93&q=init_tokens) here as well (join them together). It will allow implicitly defined tokens to be used in the task definitions as well as explicitly defined tokens.

kingpin/actors/aws/ecs.py
```diff
@@ -181,19 +174,113 @@
...
+                service_definition_file, tokens)
+            try:
+                jsonschema.validate(service_definition,
+                                    SERVICE_DEFINITION_SCHEMA)
+
+            except jsonschema.exceptions.ValidationError as e:
+                raise exceptions.InvalidOptions(e)
+
+        # Set default values.
+        service_definition.setdefault(
```
diranged: Educate me here -- whats this do?

kingpin/actors/aws/ecs.py
```diff
@@ -366,87 +422,528 @@ def _tasks_done(self, tasks):
...
+                        ...
+                    }
+                ]
+
+        Returns:
+            All containers as a flat list.
+        """
+        containers = []
+        for task_overview in task_list:
+            containers.extend(task_overview['containers'])
```
diranged: What's `extend`?

kingpin/actors/aws/ecs.py
```diff
@@ -366,87 +422,528 @@ def _tasks_done(self, tasks):
...
+        'cluster': (str, REQUIRED,
+                    'ECS cluster in the region to launch the task into.'),
+        'task_definition':
+            (str, REQUIRED,
+             'String of path to the Task Definition file template. '
+             'Must be a local file path. '
+             'Tokens to be interpolated must be of the form %VAR%.'),
+        'service_definition':
+            (str, None,
+             'String of path to the Service Definition file template.'
```
diranged: Since all these quoted strings are concatenated together, add spaces at he end of them.

kingpin/actors/aws/ecs.py
```diff
@@ -366,87 +422,528 @@ def _tasks_done(self, tasks):
...
+            self.ecs_conn.describe_services,
+            cluster=self.option('cluster'),
+            services=[service_name])
+        self._handle_failures(response['failures'], self.FAILURE_MISSING)
+
+        services = response['services']
+
+        # There should never be more than one service for a given name.
+        if len(services) > 1:
+            raise exceptions.UnrecoverableActorFailure(
```
diranged: Recoverable

kingpin/actors/aws/ecs.py
```diff
@@ -366,87 +422,528 @@ def _tasks_done(self, tasks):
...
+            **self.service_definition)
+
+        self.log.info('Creating service.')
+
+        try:
+            yield self.thread(
+                self.ecs_conn.create_service,
+                **create_parameters)
+        except Boto3Error as e:
+            raise exceptions.UnrecoverableActorFailure(e.message)
```
diranged: Recoverable

kingpin/actors/aws/ecs.py
```diff
@@ -366,87 +422,528 @@ def _tasks_done(self, tasks):
...
+            deploymentConfiguration=deployment_configuration)
+
+        self.log.info('Updating service.')
+
+        try:
+            yield self.thread(
+                self.ecs_conn.update_service,
+                **update_parameters)
+        except Boto3Error as e:
+            raise exceptions.UnrecoverableActorFailure(e.message)
```
diranged: Recoverable

kingpin/actors/aws/ecs.py
```diff
@@ -366,87 +422,528 @@ def _tasks_done(self, tasks):
...
+
+    @gen.coroutine
+    @dry('Would ensure the service is registered.')
```
diranged: This method actually shouldn't be dry -- if it is, the user will never know exactly what changes are going to be made. The underlying API call methods (`self._update_service()`, `self._create_service()`) _do_ have the `@dry` decorator, so they will properly tell the user whats going to happen.

kingpin/actors/aws/ecs.py
```diff
@@ -366,87 +422,528 @@ def _tasks_done(self, tasks):
...
+
+        if missing_count == 0 and extra_deployment_count == 0:
+            self.log.info('Service {name} deployed successfully.'.format(
+                name=service_name))
+            raise gen.Return(True)
+
+        self.log.info(
+            '{running} tasks running out of {desired}, '
+            'and {extra} deployments waiting on termination'.format(
+                running=running_count,
```
diranged: This is a tiny nit -- but you bounce around in syntax here. In other places you use `{}`, and here you use `{running}`. I have no preference, but I'd like it to be consistent at least throughout the same module.

kingpin/actors/aws/ecs.py
```diff
@@ -366,87 +422,528 @@ def _tasks_done(self, tasks):
...
+
+        # Sort events on timestamp.
+        return sorted(new_events, key=operator.itemgetter(0))
+
+    @staticmethod
+    def _arn_to_name(arn):
+        """Gets the name in an arn.
+
+        Example input:
+            arn:aws:ecs:eu-west-1:364942603424:task-definition/nextdoor-wh:135
```
diranged: Do me a favor -- remove our account ID here? :P

kingpin/actors/aws/test/test_ecs.py
```diff
@@ -24,77 +25,78 @@ def test_no_failures(self):
...
-    def test_schema_fail_extra(self):
-        with self.assertRaises(exceptions.InvalidOptions):
-            self._test(
-                {
-                    'other stuff': 0,
-                    'family': 'name',
-                    'containerDefinitions': [],
-                    'extra': 'read all about it'
-                })
+    def test_schema_ok_extra(self):
```
diranged: I do love how complete you are here... but I am willing to trust that jsonschema is doing its job. :)

diranged:
@swaggy, nice job ... a few little nits, but overall great.

kingpin/actors/aws/ecs.py
swaggy: By the way, there *was* a race condition here, and now there is code in ensure_services to handle it.

kingpin/actors/aws/ecs.py
swaggy: Sweet! Well, this method is never used outside of the class at all and is never meant to be. This is true for all the (5) staticmethods defined in these classes. So, the advantage of them being staticmethods is only that they explicitly state that they're purely functional, i.e. they have no side effects.

Originally I had them outside the class entirely but changed that recently, as this makes more sense.

kingpin/actors/aws/ecs.py
swaggy: dict.setdefault(KEY, VALUE) will set the value of KEY to VALUE iff KEY is not in the dict yet.

kingpin/actors/aws/ecs.py
swaggy: Appends the contents of a given sequence to the list.

kingpin/actors/aws/ecs.py
swaggy: :) There are spaces at the end of them!

kingpin/actors/aws/ecs.py
diranged: Just a reminder to do this in a subsequent commit..

PR #386, opened 2016-07-13
Pin coverage to version 4.0.

swaggy:
The newest version of coverage has show_missing set to false.

See: http://stackoverflow.com/questions/37733194/

We can fix with this, or by adding a .coveragerc file.


swaggy:
Abandoned in favor of #387

PR #387, opened 2016-07-13
Add .coveragerc file to show missing lines.

swaggy:
The newest version of coverage has show_missing set to false.

See: http://stackoverflow.com/questions/37733194/


mikhail:
👍

PR #391, opened 2016-07-22
Bugfix -- if no cloudformation parameters, thats fine.

diranged:


diranged:
@mikhail hey super quick review?

diranged:
Just a heads up, im gonna merge this because its a nasty bug.. if you don't use any parameters with your CF stack, you can no longer run Kingpin against it because the Parameters dict is not returned empty like we thought..

PR #392, opened 2016-08-02
Add default values to tokenized strings

mikhail:
@diranged @swaggy

### DIFF
--- a/kingpin/test/test_utils.py
+++ b/kingpin/test/test_utils.py
@@ -33,6 +33,20 @@ class TestUtils(unittest.TestCase):
         result = utils.populate_with_tokens(string, tokens)
         self.assertEquals(result, expect)
 
+    def test_populate_with_values_not_default(self):
+        tokens = {'UNIT_TEST': 'FOOBAR', 'SECOND_UNIT': 'BARBAR'}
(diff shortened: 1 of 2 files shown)


swaggy:
LGTM once we agree on a better syntax than `?`

kingpin/test/test_utils.py
```diff
@@ -33,6 +33,20 @@ def test_populate_with_env(self):
         result = utils.populate_with_tokens(string, tokens)
         self.assertEquals(result, expect)
 
+    def test_populate_with_values_not_default(self):
+        tokens = {'UNIT_TEST': 'FOOBAR', 'SECOND_UNIT': 'BARBAR'}
+        string = 'Unit %UNIT_TEST?DEFAULT% Test %SECOND_UNIT?DEFAULT2%'
```
diranged: I'd rather see us use `|` .. which means "or".. Thoughts"

mikhail: We've been discussing that for a long time... I don't really care, and i'm even ok with allowing multiple values. could be `|` or double: `||` could even be "space or space" ` or ` `?=` `:` ... don't know

diranged: I suggest just a single `|` ... using `?=` is too bash-like. A semi-colon would be OK too I suppose, but I feel like `|` really means `or` to most people.

PR #393, opened 2016-08-03
Bugfix -- we couldn't create S3 buckets in anything other than the default region.

diranged:
@Mikhail


diranged:
merging this tested and simple bugfix

PR #397, opened 2016-08-16
Make count accept a string. Boto will convert it.

swaggy:
@mikhail


mikhail:
This looks good. Let's merge it.

PR #399, opened 2016-08-18
Improve ECS service deletion.

swaggy:
This update improves logging, documentation, and script readability of the ECS service actor.
It also cleans up old task definitions.

Changes:

* Refactor from using _ensure_service.
  Previously, deletion was logging many lines with misleading text such as 'Deployed successfully'.
  Additionally, ensure_service's documentation talks about only 'updating' a service. This goes back to giving it that single responsibility.
* Clean up documentation to include the 'absent' desired state, plus other misc. documentation improvements.
* Split stop service from delete service. This allows us to add a way of only stopping and not deleting in the future, if we need.
* Add correct logging for deletion.
* Rename wait_for_deployment and is_service_deployed to wait_for_service_update and is_service_updated.
  This better correlates to their new responsibility due to service deletion.
* Adds functionality which deregisters old task definitions associated to a service that is being deleted.
  We were leaving them around before.

@mikhail @niallo

### DIFF
--- a/kingpin/actors/aws/ecs.py
+++ b/kingpin/actors/aws/ecs.py
@@ -215,7 +215,7 @@ class ECSBaseActor(base.AWSBaseActor):
     @gen.coroutine
     @dry('Would register task definition with family {0[family]}')
     def _register_task(self, task_definition):
-        """Registers a task on ECS.
+        """Registers a task.
 
         Args:
             task_definition: dict of ECS task definition parameters.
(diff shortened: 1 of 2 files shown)


kingpin/actors/aws/ecs.py
```diff
@@ -615,7 +668,7 @@ class Service(ECSBaseActor):
     """
 
     all_options = {
-        'state': (STATE, 'present', 'Desired state: present/absent'),
+        'desired_state': (STATE, 'present', 'Desired state: present/absent'),
```
diranged: Please keep this as `state` ... thats the same verb we use in all of the other state-ensuring actors, so consistency wins here IMHO.

kingpin/actors/aws/ecs.py
```diff
@@ -242,6 +242,57 @@ def _register_task(self, task_definition):
...
+        Returns:
+            List of task definitions arns matching specified restrictions.
+        """
+        self.log.info(
+            'Listing task definitions '
+            'with status {} and family prefix {}'.format(
+                status, family_prefix))
+        task_definitions = []
+        next_token = None
+        while True:
```
diranged: Can we avoid a `while True` here? Can we get a count of the pages on the first API call and then do a batch of calls to get all the data back at once?

swaggy: Looking at the API... I don't see a way.
http://boto3.readthedocs.io/en/latest/reference/services/ecs.html#ECS.Client.list_task_definitions

kingpin/actors/aws/ecs.py
swaggy: Okay, fair argument, but that's disappointing to me, since state as an option doesn't read like a verb to me. Something I can get used to though.

mikhail:
Cool! 👍

PR #400, opened 2016-08-18
ECS: Use pagination for list task definitions.

swaggy:
@diranged


diranged:
👍  when the tests pass

PR #402, opened 2016-08-19
Create a new AlertSpecBase ensurable base actor for managing RightScale AlertSpecs

diranged:
No integration tests yet .. since this actor doesn't do any subject href discovery, its hard to test it as is. It will get integration tests when we create actors that subclass from this.

### DIFF
--- a/kingpin/actors/base.py
+++ b/kingpin/actors/base.py
@@ -583,6 +583,11 @@ class EnsurableBaseActor(BaseActor):
         # Now go ahead and validate all of the user inputs the normal way
         super(EnsurableBaseActor, self).__init__(*args, **kwargs)
 
+        # Generate a list of options that will be ensured ...
+        self._ensurable_options = self.all_options.keys()
+        for option in self.unmanaged_options:
+            self._ensurable_options.remove(option)
+
         # Finally, do a class validation... make sure that we have actual
         # getter/setter methods for each of the options. This populates dicts
         # that provide references to the actual methods for execution later.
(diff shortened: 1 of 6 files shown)


diranged:
@Mikhail, @Swaggy ... review pls..

kingpin/actors/rightscale/alerts.py
```diff
@@ -370,3 +372,202 @@ def _execute(self):
...
+       }
+
+    """
+
+    all_options = {
+        'href': (str, None, 'RightScale Resource HREF to act on'),
+        'spec': (AlertSpecSchema, None, 'The actual Alert Spec definition itself.')
+    }
+    unmanaged_options = ['href']
+    desc = "AlertSpec: {spec[name]}"
```
mikhail: Haha! yes!

kingpin/actors/rightscale/alerts.py
```diff
@@ -370,3 +372,202 @@ def _execute(self):
...
+
+    @gen.coroutine
+    def _get_state(self):
+        if self.spec:
+            raise gen.Return('present')
+        raise gen.Return('absent')
+
+    @gen.coroutine
+    def _set_state(self):
+        if self.option('state') == 'absent':
```
mikhail: Missing this option above

kingpin/actors/rightscale/alerts.py
```diff
@@ -370,3 +372,202 @@ def _execute(self):
...
+        """
+        new = {}
+        desired_keys = AlertSpecSchema.SCHEMA['properties'].keys()
+        for key in desired_keys:
+            if key in spec.soul:
+                new[key] = spec.soul[key]
+        return new
+
+    @gen.coroutine
+    def _execute(self):
```
mikhail: what's the point of this `_execute` method ?

mikhail:
I left a few comments, but feel free to merge it afterwards

kingpin/actors/rightscale/alerts.py
diranged: Renamed `self.spec` to `self.existing_spec` to help clarify how `_get_state()` works.

diranged:
@Mikhail, I think this is ready for full review

kingpin/actors/base.py
```diff
@@ -597,21 +602,30 @@ def _gather_methods(self):
...
             if not self._is_method(getter) or not self._is_method(setter):
-                if option in self.unmanaged_options:
-                    continue
-
                 raise exceptions.UnrecoverableActorFailure(
                     'Invalid Actor Code Detected in %s: '
                     'Unable to find required methods: %s, %s'
                     % (self.__class__.__name__, setter, getter))
 
+            if not self._is_method(comparer):
```
mikhail: Whoa... seems convoluted. Why not go ahead and define this method and let the subclasses overwrite it? Either way need a fat docstring

diranged: The methods are all dynamic based on the actual option name. The only other way to do this is to actually change the options verifier so that it can take a getter/setter/comparer as part of the option itself. I didn't particularly like overloading that system though.

mikhail:
High level design looks good. Let's merge and we iterate on it when we start using it heavier

diranged:
Great thanks!

PR #411, opened 2016-08-30
Bugfix: Make sure we set self.changed=True after updating the runnabl…

diranged:
…e bindings.

@Mikhail, @swaggy quick review for this bugfix?


swaggy:
👍

PR #412, opened 2016-09-09
Bugfix -- handling a single alertspec on a template was broken

diranged:
This also includes some cleanup to our RightScale error text handling to
make it more obvious when we have clean rightscale error text coming
back to us.

### DIFF
--- a/kingpin/actors/rightscale/alerts.py
+++ b/kingpin/actors/rightscale/alerts.py
@@ -678,6 +678,8 @@ class AlertSpecsBase(base.EnsurableRightScaleBaseActor):
         all_resource_specs = yield self._client.find_by_name_and_keys(
             self._client._client.alert_specs, exact=True,
             subject_href=self.option('href'))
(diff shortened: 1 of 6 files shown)


diranged:
@mikhail,
  Quick review here?

kingpin/actors/rightscale/alerts.py
```diff
@@ -678,6 +678,8 @@ def _precache(self):
         all_resource_specs = yield self._client.find_by_name_and_keys(
             self._client._client.alert_specs, exact=True,
             subject_href=self.option('href'))
+        if not isinstance(all_resource_specs, list):
```
mikhail: We keep doing this. We've got to remove the `len` check inside of `find_by_name_and_keys` - it's just a bad practice to return different data for same query

mikhail:
Code is fine. My comment would require a significant rewrite of multiple sections of the code. so 👍 for now

PR #415, opened 2016-09-21
Launch max_count machines if no `count` value specified.

mikhail:
@diranged

### DIFF
--- a/kingpin/actors/rightscale/server_array.py
+++ b/kingpin/actors/rightscale/server_array.py
@@ -855,7 +855,7 @@ class Launch(ServerArrayBaseActor):
 
     Launches instances in an existing ServerArray and waits until that array
     has become healthy before returning. *Healthy* means that the array has at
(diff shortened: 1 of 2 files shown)


diranged:
This is a significant behavior change.. can you tell me what you're trying to do here?

mikhail:
This only affects the default launch count - in cases where `count` is not specified. Launching `min` value was originally chosen arbitrarily as "let's just launch the minimal needed servers." In practice if those servers are load bearing then the `min` value is only useful during least-traffic.

In the actual DSL file we've always specified `count: {MAX_COUNT}`. This is done such that transitioning traffic to a freshly-booted array does not suddenly hit a min_count array.

However... there is a huge difference between launching a count of `max_count` vs letting kingpin launch with default. The default launches *up to* the needed value, rather than new machines of that count.

diranged:
Ok, I'm fine with this.. 👍

PR #417, opened 2016-09-23
Handle ECS race condition for non-existent service

mikhail:
@swaggy @niallo 

@diranged

### DIFF
--- a/kingpin/actors/aws/ecs.py
+++ b/kingpin/actors/aws/ecs.py
@@ -742,7 +742,10 @@ class Service(ECSBaseActor):
             service_name: service name to describe.
 
         Returns:
-            Service param dict, or None if the service does not exist.
+            Service param dict.
+
+        Raises:
+            RecoverableActorFailure if number of services found is not 1.
         """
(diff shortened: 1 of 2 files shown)


swaggy:
LGTM

kingpin/actors/aws/ecs.py
```diff
@@ -753,13 +756,12 @@ def _describe_service(self, service_name):
         services = response['services']
 
         # There should never be more than one service for a given name.
-        if len(services) > 1:
+        if len(services) != 1:
```
diranged: Wait .. this means that if there is no service, it raises an exception right? Doesn't that mean that we can't create a service if its missing?

swaggy: Er, yeah, that's right. We need to catch that exception where it's called.

diranged: In that case, it shouldnt be a `RecoverableActorFailure` ... it should be a custom exception and let the caller catch that exception and handle it.

PR #422, opened 2016-10-12
Convert the aws.s3.Bucket actor to use Boto3

diranged:
Boto3 supports the NoncurrentVersionExpiration/Transition settings that
allow us to set Lifecycle policies that truly delete data even in a
versioned bucket. Boto2 simply does not support this.

Migrating to Boto3 is enough work that I decided to also leverage the
new EnsurableBaseActor class to simplify much of the logic. Now there is
no more mocking out of a bucket object, passing around of this object,
etc. Instead we just issue get/set calls directly through the Boto3 API.

Note, WIP .. Tests are still in progress.

### DIFF
--- a/kingpin/actors/aws/s3.py
+++ b/kingpin/actors/aws/s3.py
@@ -19,13 +19,11 @@
 
 import json
 import logging
-import mock
 
-from boto.s3 import lifecycle
-from boto.exception import S3ResponseError
-from boto.exception import BotoServerError
+from botocore.exceptions import ClientError, ParamValidationError
 from tornado import concurrent
 from tornado import gen
+from inflection import camelize
 import jsonpickle
 
 from kingpin.actors import exceptions
@@ -100,10 +98,19 @@ class LifecycleConfig(SchemaCompareBase):
           { "id": "unique_rule_identifier",
             "prefix": "/some_path",
             "status": "Enabled",
-            "expiration": 365,
+            "expiration": {
+              "days": 365,
+            },
+            "noncurrent_version_expiration": {
+              "noncurrent_days": 365,
+            },
             "transition": {
               "days": 90,
               "date": "2016-05-19T20:04:17+00:00",
+              "storage_class": "GLACIER",
+            },
+            "noncurrent_version_transition": {
+              "noncurrent_days": 90,
               "storage_class": "GLACIER",
             }
           }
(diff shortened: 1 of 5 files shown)


diranged:
@Mikhail, this is ready for review. This is needed so that we can properly manage lifecycles on buckets that have file versioning enabled. Also, this adds support for managing tags on buckets.

diranged:
@swaggy  you could review too btw..

kingpin/actors/aws/s3.py
```diff
@@ -334,461 +457,363 @@ def _generate_lifecycle(self, config):
...
-                self.log.debug('No bucket %s found' % self.option('name'))
-                raise gen.Return(None)
+    def _precache(self):
+        # Store a quick reference to whether or not the bucket exists or not.
+        # This allows the rest of the getter-methods to know whether or not the
+        # bucket exists and not make bogus API calls when the bucket doesn't
+        # exist.
+        buckets = yield self.thread(self.s3_conn.list_buckets)
+        matching = [
+            b for b in buckets['Buckets'] if b['Name'] == self.option('name')]
```
mikhail: Look at you using list comprehension! Bravo ;)

kingpin/actors/aws/s3.py
```diff
@@ -334,461 +457,363 @@ def _generate_lifecycle(self, config):
...
-    def _ensure_policy(self, bucket):
-        """Ensure the policy attached to the bucket is correct.
+    def _get_policy(self):
+        try:
+            raw = yield self.thread(
+                self.s3_conn.get_bucket_policy,
+                Bucket=self.option('name'))
+            exist = json.loads(raw['Policy'])
+        except ClientError as e:
+            if 'NoSuchBucketPolicy' in e.message:
```
mikhail: You're missing the `else: raise` here. Very important! Otheriwse all ClientError that is not NoScuh will be simply ignored

kingpin/actors/aws/s3.py
```diff
@@ -334,461 +457,363 @@ def _generate_lifecycle(self, config):
...
-        if self.option('logging') is not None:
-            yield self._ensure_logging(bucket)
+    def _get_tags(self):
+        try:
+            raw = yield self.thread(
+                self.s3_conn.get_bucket_tagging,
+                Bucket=self.option('name'))
+        except ClientError as e:
+            if 'NoSuchTagSet' in e.message:
+                raise gen.Return([])
```
mikhail: Ditto

diranged:
Addressed the comments.. merging after tests complete.

PR #423, opened 2016-10-18
Bugfixes around creation of new S3 buckets.

diranged:
The get_lifecycle/get_tags methods would fail if they were applied on a
resource that didn't yet exist. Oops.


diranged:
@mikhail, merging this .. do a post review, but its blocking another PR and its a clear bug..

mikhail:
yeah cool

PR #427, opened 2016-12-07
Bugfix on S3 Bucket Tag comparison.

diranged:


kingpin/actors/aws/test/test_s3.py
```diff
@@ -446,9 +447,9 @@ def test_set_lifecycle_client_error(self):
     def test_get_tags(self):
         self.actor._bucket_exists = True
         self.actor.s3_conn.get_bucket_tagging.return_value = {
-            'TagSet': []}
+            'TagSet': [{'Key': 'k1', 'Value': 'v1'}]}
```
mikhail: Please add a test where multiple key/values are in the tag set - otherwise if the loop is broken - this test still passes

mikhail:
Requested 1 change - after that 👍 to merge

kingpin/actors/aws/test/test_s3.py
diranged: will do..

PR #428, opened 2016-12-08
Strip out unicode characters in diff_dicts.

swaggy:
Note that this means it won't care about str/unicode distinction anymore in diff.

I think we're okay with that, but I noted it in the docstring.

@diranged


diranged:
So I think this is fine -- but I have no doubt that down the road we'll revisit this and find a smarter way to do it.

PR #431, opened 2016-12-26
Compare S3 bucket tags with some intelligence.

diranged:
We can't simply use the logic in the base class compare_xxx() method
because it takes into account order in lists, which don't matter with
tags. Instead we need a dedicated compare_tags() method that understands
this.

### DIFF
--- a/kingpin/actors/aws/s3.py
+++ b/kingpin/actors/aws/s3.py
@@ -830,6 +830,26 @@ class Bucket(base.EnsurableAWSBaseActor):
         raise gen.Return(tagset)
 
     @gen.coroutine
+    def _compare_tags(self):
+        new = self.option('tags')
+        if new is None:
+            self.log.debug('Not managing Tags')
+            raise gen.Return(True)
+
+        exist = yield self._get_tags()
+
+        diff = self._diff_dicts(exist, new)
+        if not diff:
+            self.log.debug('Bucket tags match')
+            raise gen.Return(True)
+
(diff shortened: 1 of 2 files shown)


kingpin/actors/aws/s3.py
```diff
@@ -830,6 +830,29 @@ def _get_tags(self):
...
     @gen.coroutine
+    def _compare_tags(self):
+        new = self.option('tags')
+        if new is None:
+            self.log.debug('Not managing Tags')
+            raise gen.Return(True)
+
+        exist = yield self._get_tags()
+
+        # Now, diff our new policy from the existing policy. If there is no
```
mikhail: these comments are useless

kingpin/actors/aws/s3.py
```diff
@@ -830,6 +830,29 @@ def _get_tags(self):
...
+        # difference, then we bail out of the method.
+        diff = self._diff_dicts(exist, new)
+        if not diff:
+            self.log.debug('Bucket tags match')
+            raise gen.Return(True)
+
+        # Now, print out the diff..
```
mikhail: ditto

kingpin/actors/aws/s3.py
```diff
@@ -830,6 +830,29 @@ def _get_tags(self):
         raise gen.Return(tagset)
 
     @gen.coroutine
+    def _compare_tags(self):
```
mikhail: Where are you using this function?

PR #435, opened 2017-02-07
Pass Kingpin ECS actor ENV vars to task definition

nehalrp:


kingpin/actors/aws/ecs.py
```diff
@@ -488,7 +499,8 @@ def __init__(self, *args, **kwargs):
         super(RunTask, self).__init__(*args, **kwargs)
         self.task_definition = self._load_task_definition(
```
diranged: This looks great, but lets also apply this to the `self.service_definition` too for consistency.

PR #436, opened 2017-02-14
Add more logging to Kingpin to help debug rare ECS race condition.

swaggy:
@diranged @mikhail @niallo 

About once every two weeks we run into a case where the actor hangs on updating primary deployment. I've made the same API calls at those times and looked in the ECS console - everything looks fine. And yet it's stuck.

I don't like having to do this, but I need some more information. This'll print out more local variables and the exception message, which'll help me get to the bottom of this bug.


diranged:
👍

PR #437, opened 2017-03-23
ECS: Optionally reuse existing count when updating a service.

swaggy:
@mikhail @diranged


mikhail:
@swaggy looking good except that you need to update the documentation twice. Namely in the docstring

swaggy:
Verbal OK from Mikhail

PR #438, opened 2017-03-27
Support python3.

mikhail:
Starting this to track changes. cc @ashanbrown

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 build/
 docs/.ropeproject
 .DS_Store
+build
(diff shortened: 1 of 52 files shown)


ashanbrown:
@diranged I've almost got this working.  I don't really know enough about how kingpin works to fix test_server_template.TestServerTemplateActor.test_precache, which I am currently disabling.  Several tests were failing in python3 because we were missing sufficient entries in the `side_effect` to handle all the calls.  On python2, raising StopIteration would just cause the test to terminate early skipping the rest of the test, but on python3 a StopIteration does not cause a generate to stop (see https://www.python.org/dev/peps/pep-0479/ for details).  FWIW, you can backport this behavior to python2 with:

```from __future__ import generator_stop```

but you'd have to do it on every module.

PR #447, opened 2017-07-03
Paginate the IAM get_all_entities() method call

### LINKED ISSUE #446: Once you have more than a certain number of IAM entities, we don't paginate...
The code https://github.com/Nextdoor/kingpin/blob/master/kingpin/actors/aws/iam/entities.py#L339-L343 doesn't paginate, so once you have a certain number of IAM entities, the code always thinks that one of them is missing and tries to recreate it.

diranged:
If you have more than 100 iam entities, by default you have to paginate
the results. Build in pagination, but also set the max_items to 1000
(API limit) so that we can get all our results back in one API call.

This closes #446

### DIFF
--- a/kingpin/actors/aws/iam/entities.py
+++ b/kingpin/actors/aws/iam/entities.py
@@ -42,6 +42,11 @@ __author__ = 'Matt Wise <matt@nextdoor.com>'
 # do this.
 EXECUTOR = concurrent.futures.ThreadPoolExecutor(10)
 
+# The maximum number of items returned to us in a get_all_*/list_* api call.
+# Defaults to 100, but setting to 1000 to reduce the number of API calls we
+# make.
+MAX_ITEMS = 1000
+
 
 class EntityBaseActor(base.IAMBaseActor):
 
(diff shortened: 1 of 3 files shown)


kingpin/actors/aws/iam/entities.py
```diff
@@ -335,34 +340,44 @@ def _get_entity(self, name):
...
-        if len(entity) > 1:
-            raise exceptions.RecoverableActorFailure(
-                'More than one %s found matching %s! Am I crazy?!' %
-                (self.entity_name, name))
+            # Check our result for the entity.. if its there, great.
+            # Otherwise, we'll move on.
+            entity = [entity for entity in result['%ss' % self.entity_name]
+                      if entity['%s_name' % self.entity_name] == name]
+
+            if len(entity) > 0:
```
stlava: Why did you remove the `len(entity) > 1` check/exception and change it to just return the first entity?

diranged: That check was always bogus - Amazon will never let you have two Roles, Users or Groups with the same name.

PR #449, opened 2017-07-07
Fix Group Actor Context Token Substitution

### LINKED ISSUE #448: Group Actors Fill Context Tokens Before Loading Context
```yaml
# test.yaml
actor: group.Sync
options:
  contexts: "test-context.yaml"
  acts:
    - actor: misc.Note
      options:
        message: "{CONTEXT_VALUE|Default Value}"
```
```yaml
# test-context.yaml
- CONTEXT_VALUE: "Hello World"
```

This outputs `"Default Value"` instead of `"Hello World"`.

It appears that when a group actor is initialized, it will attempt to fill in context tokens _prior to loading any specified context_ as part of the `BaseActor` constructor (which calls `_fill_in_contexts` with the `_init_context`).

This appears to be related to a larger issue, caused by multiple levels of group actors, each with their own context, where tokens are removed too soon (because they aren't defined until later `contexts`). An example (this prints `"Foo Bar"` not `"Hello World"`):

```yaml
# test2.yaml
actor: group.Sync
options:
  contexts: "test-context1.yaml"
  acts:
    - actor: group.Sync
       options:
         contexts: "test-context2.yaml"
         acts:
           - actor: misc.Note
              options:
                message: "{HELLO_TOKEN|Foo} {WORLD_TOKEN|Bar}"
```
```yaml
# test-context1.yaml
- HELLO_TOKEN: "Hello"
```
```yaml
# test-context2.yaml
- WORLD_TOKEN: "World"
```

I'll submit a PR with a potential solution for both. This will change how context tokens are filled. Instead of blindly replacing all context tokens in the options subtree (via converting the JSON to a string), it'll traverse the tree and stop substituting context tokens in any group actors which defines a `contexts` field. The in-progress PR is here: https://github.com/Nextdoor/kingpin/compare/master...colinking:colin/group_contexts

colinking:
Fixes #448.

This PR fixes a problem where context tokens would be filled before all relevant contexts were loaded. Context tokens are now only filled after all parent actor contexts have been loaded. See #448 for examples.

cc @diranged @mikhail

### DIFF
--- a/kingpin/actors/base.py
+++ b/kingpin/actors/base.py
@@ -135,11 +135,8 @@ class BaseActor(object):
             timeout: (Str/Int/Float) Timeout in seconds for the actor.
         """
         self._type = '%s.%s' % (self.__module__, self.__class__.__name__)
-        self._options = options
-        self._desc = desc
         self._dry = dry
         self._warn_on_failure = warn_on_failure
-        self._condition = condition
         self._init_context = init_context
         self._init_tokens = init_tokens
 
@@ -147,10 +144,16 @@ class BaseActor(object):
         if timeout is None:
             self._timeout = self.default_timeout
 
+        # Fill the context into the description, condition, and options.
+        actor = self._type.replace('kingpin.actors.','')
+        act = {'actor': actor, 'desc': desc, 'condition': condition, 'options': options}
         # strict about this -- but in the future, when we have a
         # runtime_context object, we may loosen this restriction).
-        self._fill_in_contexts(context=self._init_context,
-                               strict=self.strict_init_context)
+        updated_act = self._fill_in_contexts(act, context=self._init_context,
+                                             strict=self.strict_init_context)
+        self._desc = updated_act['desc']
+        self._condition = updated_act['condition']
+        self._options = updated_act['options']
 
         self._setup_log()
         self._setup_defaults()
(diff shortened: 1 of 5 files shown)


kingpin/actors/base.py
```diff
@@ -373,65 +381,88 @@ def _check_condition(self):
...
+        # Inject contexts into the description.
+        updated_desc = self._fill_contexts_in_string(act.desc, context, strict)
+
+        # Inject contexts into the condition.
+        updated_condition = self._fill_contexts_in_string(act.condition,
+                                                          context, strict)
+
+        # Inject contexts into the options, skipping over any group sub-actors
+        # which declare more context tokens.
+        if actors.group == sys.modules[get_actor_class(act.actor).__module__]:
```
mikhail: Refactor this equality into a named variable. Maybe `actor_is_group` ?

kingpin/actors/test/test_group.py
```diff
@@ -191,6 +191,13 @@ def test_build_action_group(self):
         ret = actor._build_action_group({'TEST': 'CONTEXT'})
         self.assertEquals(ret[0]._init_context, {'TEST': 'CONTEXT'})
 
+    def test_fill_contexts_in_act_invalid_context(self):
```
mikhail: Where is the actual test that nested group actors do not clobber default contexts?

colinking: See the most recent commit -- my bad that they weren't included in the original PR.

PR #453, opened 2017-12-14
CF Actor - Add enable_termination_protection option

nehalrp:
@diranged Still need to finish updating docs and tests, but I wanted to get your eyes on the implementation here.

### DIFF
--- a/kingpin/actors/aws/cloudformation.py
+++ b/kingpin/actors/aws/cloudformation.py
@@ -118,12 +118,26 @@ class OnFailureConfig(StringCompareBase):
     """Validates the On Failure option.
 
     The `on_failure` option can take one of the following settings:
-      `DO_NOTHING`, `ROLLBACK`, `DELETE`
+    `DO_NOTHING`, `ROLLBACK`, `DELETE`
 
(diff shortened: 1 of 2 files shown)


stlava:
@nehalrp do you know if this also puts termination protection on resources under the stack as well or to the stack itself?

nehalrp:
@stlava It's purely on the stack itself.

nehalrp:
@diranged Updated with tests.

kingpin/actors/aws/cloudformation.py
```diff
@@ -762,6 +781,9 @@ class Stack(CloudFormationBaseActor):
         'timeout_in_minutes': (int, 60,
                                'The amount of time that can pass before the '
                                'stack status becomes CREATE_FAILED'),
+        'enable_termination_protection': (TerminationProtectionConfig, None,
```
diranged: Add to the docstrings above fro this option - also while you're there, can you include the `role_arn` docs too?

nehalrp: Ah, missed those particular ones.

PR #454, opened 2017-12-19
Revert "CF Actor - Add enable_termination_protection option (#453)"

nehalrp:
This reverts commit 7d9ecec57e4434bc70831c552fe7f8e5f908c279.

There's an interesting bug whereby when running the `Stack` actor via the `Group` actor explicitly (either sync or async) and not defining the `enable_termination_protection` option it defaults to `None` and so botocore throws a validation error as it is a `bool` parameter. However, when ran the actor is ran directly (without being explicitly defined under a Group actor) it defaults to `False` and runs file.

NOTE: It's interesting that the Group actor is actually used (`[kingpin.actors.group.Sync] Beginning 2 actions`) when not explicitly defined in the file, but it still defaults properly and runs properly. This leads me to believe there may be some issue with how our options are set when trying to run through different `Group` contexts from a file...

### DIFF
--- a/kingpin/actors/aws/cloudformation.py
+++ b/kingpin/actors/aws/cloudformation.py
@@ -118,26 +118,12 @@ class OnFailureConfig(StringCompareBase):
     """Validates the On Failure option.
 
     The `on_failure` option can take one of the following settings:
-    `DO_NOTHING`, `ROLLBACK`, `DELETE`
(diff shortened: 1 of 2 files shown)


diranged:
Before you revert this - check in with @Mikhail and see if he can help this morning.

PR #458, opened 2018-01-26
Handle other parts of service definition indiscriminately.

swaggy:
Instead of only passing along specfic parts, pass everything except for some specific parts that aren't allowed to change.
This'll allow us to more easily keep up with ECS feature changes.

### DIFF
--- a/kingpin/actors/aws/ecs.py
+++ b/kingpin/actors/aws/ecs.py
@@ -370,7 +370,7 @@ class ECSBaseActor(base.AWSBaseActor):
 
     @staticmethod
     def _load_service_definition(
-            service_definition_file, tokens, default_tokens={}):
+            service_definition_file, tokens, default_tokens=None):
         """Loads and verifies a service definition template file, and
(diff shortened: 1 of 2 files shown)


kingpin/actors/aws/ecs.py
```diff
@@ -981,12 +976,15 @@ def _update_service(self, service_name, existing_service, override=None):
...
-        deployment_configuration = self.service_definition[
-            'deploymentConfiguration']
         update_parameters = dict(
             cluster=self.option('cluster'),
             service=service_name,
-            deploymentConfiguration=deployment_configuration)
+            **self.service_definition)
+
+        if 'loadBalancers' in update_parameters:
+            del update_parameters['loadBalancers']
```
mikhail: If you care... `update_parameters.pop("loadBalancers", 1)` -- no need for the if statement.

swaggy: oh nice

PR #464, opened 2018-04-05
ECS: Stop setting defaults for immutable fields.

swaggy:
If they aren't supplied, amazon provides a default.


pjanakiraman:
LGTM.

PR #465, opened 2018-06-30
Integrate an API Call Queue for cf.describe_stacks.

swaggy:
This will serialize all the requests to the CloudFormation describe_stacks API to better protect against rate limits.

This is a pretty complicated implementation, and I don't like how it integrates with testing at the moment.
Each test creates a new ioloop, breaking the existing ApiCallQueues, as the process_queue callback will not be re-entered. Haven't figured out a way around this yet other than clearly the ApiCallQueues between tests.

### DIFF
--- a/kingpin/actors/aws/api_call_queue.py
+++ b/kingpin/actors/aws/api_call_queue.py
@@ -1,0 +1,133 @@
+from boto import exception as boto_exception
+from botocore import exceptions as botocore_exceptions
+from tornado import concurrent
+from tornado import gen
+from tornado import queues
+from tornado import ioloop
+
+from kingpin.actors import exceptions
+
+EXECUTOR = concurrent.futures.ThreadPoolExecutor(10)
+
+
+class ApiCallQueue:
+    """
+    Handles queueing up and sending AWS api calls serially,
+    with exponential backoff when there is throttling.
+
+    Supports both boto2 and boto3.
+    """
+
+    def __init__(self):
+        self.executor = EXECUTOR
+
+        self._queue = queues.Queue()
+        ioloop.IOLoop.current().spawn_callback(self._process_queue)
+
+        # Used for controlling how fast the work queue is processed,
+        # with exponential delay on throttling errors.
+        self.delay_min = 0.25
+        self.delay_max = 30
+        # We don't have a delay until we first get throttled.
+        self.delay = 0
+
+        # There are a number of different rate limiting messages
+        # boto2 can return when rate limits are reached, depending
+        # on which apis are used.
+        self.boto2_throttle_strings = (
+            'Throttling',
+            'Rate exceeded',
+            'reached max retries',
+        )
+
+    @gen.coroutine
+    def call(self, api_function, *args, **kwargs):
+        result_queue = queues.Queue(maxsize=1)
+        yield self._queue.put((result_queue, api_function, args, kwargs))
+        result = yield result_queue.get()
+        if isinstance(result, Exception):
+            raise result
+        raise gen.Return(result)
+
+    @gen.coroutine
+    def _process_queue(self):
+        while True:
+            result_queue, api_function, args, kwargs = yield self._queue.get()
+            try:
+                result = yield self._call(api_function, *args, **kwargs)
+            except BaseException as e:
+                result = e
+            yield result_queue.put(result)
+            if self.delay > 0:
+                yield gen.sleep(self.delay)
+
+    @gen.coroutine
+    def _call(self, api_function, *args, **kwargs):
+        while True:
+            try:
+                result = yield self._thread(
+                    api_function, *args, **kwargs)
+                self.decrease_delay()
+                raise gen.Return(result)
+            except boto_exception.BotoServerError as e:
+                # Boto2 exception.
+                if e.error_code in self.boto2_throttle_strings:
+                    self.increase_delay()
+                    yield gen.sleep(self.delay)
+                    continue
+
+                # If we're using temporary IAM credentials, when those expire
+                # we can get back a blank 400 from Amazon. This is confusing,
+                # but it happens because of
+                # https://github.com/boto/boto/issues/898.
+                # In most cases, these temporary IAM creds can be re-loaded by
+                # reaching out to the AWS API (for example, if we're using an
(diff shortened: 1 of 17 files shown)


swaggy:
Note: Haven't written unit tests for the ApiCallQueue itself yet - will do so Monday.

swaggy:
@diranged This is ready for review - Cleaned it up and added tests.

kingpin/actors/aws/api_call_queue.py
```diff
@@ -0,0 +1,109 @@
...
+        # boto2 can return when rate limits are reached, depending
+        # on which apis are used.
+        self.boto2_throttle_strings = (
+            'Throttling',
+            'Rate exceeded',
+            'reached max retries',
+        )
+
+    @gen.coroutine
+    def call(self, api_function, *args, **kwargs):
```
diranged: Add docstrings to all methods please.. especially the public ones. :)

kingpin/actors/aws/api_call_queue.py
```diff
@@ -0,0 +1,109 @@
...
+EXECUTOR = concurrent.futures.ThreadPoolExecutor(10)
+
+
+class ApiCallQueue:
+    """
+    Handles queueing up and sending AWS api calls serially,
+    with exponential backoff when there is throttling.
+
+    Supports both boto2 and boto3.
+    """
```
diranged: Can you add documentation in here on how to use this in practice?

kingpin/actors/aws/api_call_queue.py
```diff
@@ -0,0 +1,109 @@
...
+            'Throttling',
+            'Rate exceeded',
+            'reached max retries',
+        )
+
+    @gen.coroutine
+    def call(self, api_function, *args, **kwargs):
+        result_queue = queues.Queue(maxsize=1)
+        yield self._queue.put((result_queue, api_function, args, kwargs))
+        result = yield result_queue.get()
```
diranged: Is there a timeout here on this at all?

kingpin/actors/aws/api_call_queue.py
```diff
@@ -0,0 +1,109 @@
...
+                result = e
+            yield result_queue.put(result)
+            yield gen.sleep(self.delay)
+
+    @gen.coroutine
+    def _call(self, api_function, *args, **kwargs):
+        while True:
+            try:
+                result = yield self._thread(
+                    api_function, *args, **kwargs)
```
diranged: This should fit on one line?

kingpin/actors/aws/api_call_queue.py
```diff
@@ -0,0 +1,109 @@
...
+            except botocore_exceptions.ClientError as e:
+                # Boto3 exception.
+                if e.response['Error']['Code'] == 'Throttling':
+                    self.increase_delay()
+                    yield gen.sleep(self.delay)
+                else:
+                    self.decrease_delay()
+                    raise e
+
+    def decrease_delay(self):
```
diranged: Definitely document all of this..

kingpin/actors/aws/base.py
```diff
@@ -183,18 +182,55 @@ def __init__(self, *args, **kwargs):
...
+        The API function is assumed to be a synchronous function.
+        It will be run on a concurrent thread using run_on_executor.
+
+        The queue_identifier argument specifies which queue to use.
+        If the queue doesn't exist, it will be created.
+
+        Example:
+            >>> zones = yield api_call_with_queueing(ec2_conn.get_all_zones)
+        """
+        if queue_name not in NAMED_API_CALL_QUEUES:
```
diranged: `queue_name` is not in your example code 3 lines above..

kingpin/actors/aws/base.py
```diff
@@ -183,18 +182,55 @@ def __init__(self, *args, **kwargs):
...
+            NAMED_API_CALL_QUEUES[queue_name] = (
+                api_call_queue.ApiCallQueue())
+        queue = NAMED_API_CALL_QUEUES[queue_name]
+        try:
+            result = yield queue.call(api_function, *args, **kwargs)
+        except Exception as e:
```
diranged: Can we limit the scope here at all?

kingpin/actors/aws/cloudformation.py
```diff
@@ -316,8 +316,10 @@ def _get_stack(self, stack):
             <Stack Dict> or <None>
         """
         try:
-            stacks = yield self.thread(self.cf3_conn.describe_stacks,
-                                       StackName=stack)
+            stacks = yield self.api_call_with_queueing(
+                self.cf3_conn.describe_stacks,
+                queue_name='describe_stacks',
```
diranged: To simplify this - if the intention is for the queue to always be unique for each type of api call, why not just dynamically generate the queue name based on the api_function name?

swaggy: I would much prefer that - but I couldn't figure out how to actually do that in practice. Ideas?

diranged: `my_function.__name__`?

kingpin/actors/aws/api_call_queue.py
swaggy: No, and that's by design.

If an API call is being rate limited, this will back off the attempts and try forever until it stops getting rate limited. The API calls themselves will have some timeout from boto, which'll cause an exception to raise, and the Actors have their own run timeout that wraps this.

It may make sense to add something like a 5 minute timeout here, in case we are getting endlessly rate limited, but why? I feel like trying continuously with backoff for the rate limit case is a fine way to handle rate limiting.

It could use some logging when it does get rate limited, though.

kingpin/actors/aws/cloudformation.py
swaggy: Will give that a shot. That was my initial attempt as well, but it didn't work there because the scope was larger - I was changing all boto function calls rather than just `describe_stacks`, so it caused failures in many tests because parallel calls were now serialized. Maybe it'll work with just `describe_stacks`.

swaggy: Right, so the problem is that a `Mock` doesn't have a `__name__`, so it fails when the api calls are mocked out in tests.

swaggy:
This is ready for re-review.

PR #472, opened 2019-03-06
Filter out ResolvedValue from the params

stlava:
When using SSM parameters the resolved values are returned for the stack. They can't be compared as the local params aren't resolved (obviously).

### DIFF
--- a/kingpin/actors/aws/cloudformation.py
+++ b/kingpin/actors/aws/cloudformation.py
@@ -896,7 +896,7 @@ class Stack(CloudFormationBaseActor):
         # Get and compare the parameters we have vs the ones in CF. If they're
         # different, plan to do an update!
         if self._diff_params_safely(
-                stack.get('Parameters', {}),
+                stack.get('Parameters', []),
                 self._parameters):
             needs_update = True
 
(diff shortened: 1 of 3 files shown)


kingpin/actors/aws/cloudformation.py
```diff
@@ -959,6 +959,18 @@ def _diff_params_safely(self, remote, local):
             remote = [pair for pair in remote if pair['ParameterKey'] != p]
             local = [pair for pair in local if pair['ParameterKey'] != p]
 
+        # Remove any resolved parameter values that were inserted by SSM
+        # so that only supplied parameter values are compared.
+        filtered_remote = []
+        for param in remote:
+            filtered_param = {}
+            for k, v in param.items():
```
nehalrp:
```
if `ResolvedValue` in param:
    del param[`ResolvedValue`]
```

Or is mutating the dict here a problem?

stlava: As a rule of thumb I don't like modifying the inputs.

nehalrp: 👍

PR #481, opened 2019-12-11
Log status reason on stack and change set failures

smmorneau:
Display status reasons to clarify errors and improve debugging during kingpin runs.

`describe_stacks` returns a `StackStatusReason`: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-describing-stacks.html#using-cfn-describing-stacks-describe-stacks

`describe-change-set` returns a `StatusReason`: https://docs.aws.amazon.com/cli/latest/reference/cloudformation/describe-change-set.html#examples


LaikaN57:
Integrated into #487

PR #482, opened 2019-12-13
Fix changeset error log

smmorneau:
Fixes KeyError.

Follow up to https://github.com/Nextdoor/kingpin/pull/481

PR #483, opened 2019-12-13
Fix missing status reason logging for CFN actor

LaikaN57:
Fixes KeyError.

Follow up to #481

Adding on to #482


LaikaN57:
@smmorneau Found another place we should add this fix.

PR #484, opened 2019-12-13
Revert "Log status reason on stack and change set failures"

LaikaN57:
Reverts Nextdoor/kingpin#481


LaikaN57:
Integrated into #487

PR #486, opened 2020-01-10
Cancel cfn update on timeout

RKTMN:
When we get into a state where our CloudFormation stack is unable to update, the Kingpin actor will time out. However, the stack then has to be cancelled (and therefore rolled back) manually, and this manual step has been missed several times before. 

This change is intended so that when the CloudFormation actor in particular times out, it will also cancel the stack update.

### DIFF
--- a/kingpin/actors/base.py
+++ b/kingpin/actors/base.py
@@ -378,7 +378,7 @@ class BaseActor(object):
 
         Parses the objects self._options dict (by converting it into a JSON
         string, substituting, and then turning it back into a dict) and the
-        self._desc string and replaces any {KEY}s with the valoues from the
+        self._desc string and replaces any {KEY}s with the values from the
         context dict that was supplied.
 
         Args:
@@ -478,7 +478,7 @@ class BaseActor(object):
         """
         self.log.debug('Beginning')
 
-        # Any exception thats raised by an actors _execute() method will
+        # Any exception that's raised by an actors _execute() method will
         # automatically cause actor failure and we return right away.
         result = None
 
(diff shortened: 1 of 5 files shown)


kingpin/actors/base.py
```diff
@@ -488,7 +488,10 @@ def execute(self):
             raise gen.Return()
 
         try:
-            result = yield self.timeout(self._execute)
+            if hasattr(self, '_execution_timeout'):
```
diranged: I do not want to muddy up the base actor with information that is relevant only for a particular type of actual actor - ie, `_execution_timeout` is a method only defined in the cloudformation.py file, so the `base.py` file should know nothing about it.

kingpin/actors/base.py
```diff
@@ -488,7 +488,10 @@ def execute(self):
...
+                result = yield self._execution_timeout(self._execute)
```
diranged: Wrapping the update-timeout checking at this level is problematic because it could actually lead to the stack update being canceled before the latest update is even tried. Here's the workflow:

1. Bob issues an stack update at 12:05PM and the stack goes into an UPDATING state...
2. Joe calls Kingpin to update the stack at 12:20PM with a 5M timeout.. but doesn't know anything about Bob doing that work.
3. Kingpin immediately starts a 5M timer and THEN checks the stack's current state (with the `ensure_state()` method). It sits on the stack while it finishes its UPDATING from Bob before it ever begins the update for Joe.
4. Bob's update takes a while and goes on to 12:25PM - at which point Joe's Kingpin instance issues a CANCEL call to the stack update.

See the issue?

PR #491, opened 2020-03-17
Fix Snyk Integration

LaikaN57:
Changing line 33 to `python-rightscale @ https://github.com/diranged/python-rightscale-1/tarball/nextdoor#egg=python-rightscale` will save the egg using the same names as the lib's name, which will allow Snyk to identify and link the library with its sources.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,7 +30,7 @@ retrying
 #
 # NOTE: Version specified in the setup.py file
 # NOTE: mock is used inside the rightscale actor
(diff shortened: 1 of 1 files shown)


LaikaN57:
From ticket: ```
When it's enabled, we want to show you only new issues being introduced by the PR. For that, we need to test your both your source branch and target branch and compare the found issues. Since your target branch (master) always ends with an error, we can't finish the check.

If you merge PR 491 to master, future commit status checks will work. Verified this on my fork.
```

PR #493, opened 2020-04-01
Extra checks for ServiceNotActiveException

RKTMN:
Helps to diagnose Kingpin failures like the following: https://build.corp.nextdoor.com/job/prod-nextdoor-deploy-au1/4968/console

### DIFF
--- a/kingpin/actors/aws/ecs.py
+++ b/kingpin/actors/aws/ecs.py
@@ -32,7 +32,7 @@ from kingpin.constants import REQUIRED, STATE
 
 log = logging.getLogger(__name__)
 
-__author__ = 'Steve Mostovoy <smostovoy@nextdoor.com>'
+__author__ = 'Willem Van Eck <willem@nextdoor.com>'
 
 
 class ECSAPIException(exceptions.RecoverableActorFailure):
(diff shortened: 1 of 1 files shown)


kingpin/actors/aws/ecs.py
```diff
@@ -32,7 +32,7 @@
 
 log = logging.getLogger(__name__)
 
-__author__ = 'Steve Mostovoy <smostovoy@nextdoor.com>'
```
RKTMN, line 35: We might as well have exceptions and stack traces routed to a current Nextdoorian...

kingpin/actors/aws/ecs.py
```diff
@@ -1044,11 +1044,30 @@ def _update_service(self, service_name, existing_service, override=None):
...
+            if not service_is_active:
+                # We can only update an existing service.
+                self.log.error(
+                    'Could not find service with name {} to update '
+                    'in {}. Update is likely to fail.'.format(
+                        service_name, self._format_location()))
+            yield self.api_call(
+                self.ecs_conn.update_service,
+                **update_parameters)
+        except self.ecs_conn.exceptions.ServiceNotFoundException as e:
```
RKTMN, line 1063: My best guess at how to catch the error, from the discussion here: https://nextdoor.slack.com/archives/C03H4GRDF/p1585779522133400

kingpin/actors/aws/ecs.py
```diff
@@ -1044,11 +1044,30 @@ def _update_service(self, service_name, existing_service, override=None):
...
+        
+        
         self.log.info('Updating service.')
-
-        yield self.api_call(
-            self.ecs_conn.update_service,
-            **update_parameters)
+        try:
+            self.log.info('Checking if existing service is active.')
+            service_is_active = (existing_service and
```
RKTMN, line 1052: Checking if service being inactive is really the state its in before the deployment fails (it should be, based on AWS docs): https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_UpdateService.html

kingpin/actors/aws/ecs.py
```diff
@@ -1045,7 +1045,25 @@ def _update_service(self, service_name, existing_service, override=None):
             update_parameters.update(override)
 
         self.log.info('Updating service.')
-
+        self.log.info('Checking if existing service is active...')
+        service_is_active = (existing_service and
```
davidkern, line 1049: Better to validate `existing_service` parameter is not None (and throw an exception) at the top of the function, and then assume it is ok for the rest of the function body.

PR #496, opened 2020-05-07
S3 Lifecycle Policy: Add Filter Option

LaikaN57:
AWS added a new API to change some of the functionality slightly. You can now use Filter>Prefix instead of just Prefix to add multiple lifecycle policy rules onto the same prefixes. Not quite sure how it determines which policy wins but I will leave that as an exercise for the reader.

### DIFF
--- a/kingpin/actors/aws/s3.py
+++ b/kingpin/actors/aws/s3.py
@@ -149,8 +149,7 @@ class LifecycleConfig(SchemaCompareBase):
     .. code-block:: json
 
         [
-          {
-            "id": "unique_rule_identifier",
+          { "id": "unique_rule_identifier",
             "filter": {
               "prefix": "/some_path"
             },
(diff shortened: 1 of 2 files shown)


kingpin/actors/aws/s3.py
```diff
@@ -521,6 +580,14 @@ def _generate_lifecycle(self, config):
         for c in config:
             self.log.debug('Generating lifecycle rule from foo: %s' % c)
 
+            # You must supply at least 'prefix' or 'filter' in your
+            # lifecycle config. This is tricky to check in the jsonschema, so
+            # we do it here.
+            if not any(k in c for k in ('prefix', 'filter')):
```
diranged, line 586: You can do this with oneOf... https://stackoverflow.com/questions/24023536/how-do-i-require-one-field-or-another-or-one-of-two-others-but-not-all-of-them

kingpin/actors/aws/s3.py
```diff
@@ -178,48 +178,117 @@ class LifecycleConfig(SchemaCompareBase):
         'uniqueItems': True,
         'items': {
             'type': 'object',
-            'required': ['id', 'prefix', 'status'],
+            'allOf': [
```
ivanpeng, line 181: Please update the documentation above.

kingpin/actors/aws/s3.py
```diff
@@ -178,48 +178,117 @@ class LifecycleConfig(SchemaCompareBase):
...
-                            'pattern': '^[0-9]+$',
-                        },
-                        'date': {
+                        'prefix': {
                             'type': 'string',
-                            'format': 'date-time',
                         },
-                        'expired_object_delete_marker': {
-                            'type': 'boolean',
+                        'tag': {
```
ivanpeng, line 241: You can probably move this out to a tag object, and reference it there, since you're playing with python dicts. That way you can also reuse this down in the other blocks

diranged: Good call - there's actually a proper syntax in JsonSchema where you create a Tag object and just refer to it by name. You move this into the `definitions` key at the top of the schema. Then you can use `$ref` to refer to it... kinda like this:
```
"tag": { "$ref": "#/definitions/Tag" }
```

LaikaN57: yup saw that, learning json schema on the fly :) I will switch to using a definition

LaikaN57:
After much fighting with the tester, schema validator, linter, and document generator... I think I have a working solution.

PR #497, opened 2020-05-11
S3 LCP: Fix Transitions API Key

LaikaN57:
Fixes:
```
19:58:23   CRITICAL  [S3 Bucket] Invalid Lifecycle Configuration: Parameter validation failed:
Unknown parameter in LifecycleConfiguration.Rules[0]: "Transition", must be one of: Expiration, ID, Prefix, Filter, Status, Transitions, NoncurrentVersionTransitions, NoncurrentVersionExpiration, AbortIncompleteMultipartUpload
```

Here are the API Changes when we switched to using `put_bucket_lifecycle_configuration `:
- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.put_bucket_lifecycle
- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.put_bucket_lifecycle_configuration

### DIFF
--- a/kingpin/actors/aws/test/test_s3.py
+++ b/kingpin/actors/aws/test/test_s3.py
@@ -30,10 +30,10 @@ class TestBucket(testing.AsyncTestCase):
                     'prefix': '/test',
                     'status': 'Enabled',
                     'expiration': '30',
-                    'transition': {
+                    'transitions': [{
                         'days': 45,
                         'storage_class': 'GLACIER',
-                    }
+                    }]
                 }],
                 'logging': {
                     'target': 'test_target',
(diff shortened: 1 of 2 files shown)


kingpin/actors/aws/test/test_s3.py
```diff
@@ -30,10 +30,10 @@ def setUp(self):
                     'prefix': '/test',
                     'status': 'Enabled',
                     'expiration': '30',
-                    'transition': {
+                    'transitions': [{
```
diranged, line 33: @LaikaN57 Does this mean you broke the API for people using `transition: ...`?

LaikaN57: Yup this broke with this change: https://github.com/Nextdoor/kingpin/pull/496/files#diff-8782cedb3640968fdf2505ccfc4f9d81R938

PR #502, opened 2020-05-19
Escape base actor options

stlava:


LaikaN57:
@stlava Maybe add some documentation to this effect as well.

PR #504, opened 2020-06-17
Upgrade to python3

stlava:
- Update code be python3 compatible
- Updates to tornado 6
- Cleans up dependencies
- Replaces copied in parts of `tornado_rest_client` with actual module


.travis.yml
```diff
@@ -1,12 +1,12 @@
 language: python
 python:
-  - "2.7"
+  - "3.7"
```
diranged, line 3: Are you specifically breaking 2.x support?

stlava: Yes, I’m excluding python 2 in the list of supported versions under setup.py and creating a new major version of kingpin to signify there is a backwards incompatible change.

PR #507, opened 2020-08-21
Add support CFN templates that are located in S3

mrterry:


kingpin/actors/aws/cloudformation.py
```diff
@@ -1230,3 +1238,24 @@ def _execute(self):
...
+
+
+def parse_s3_url(url):
+    """ Gets bucket name, region, and key from url
+
+    returns bucket name, region, key
+    """
+    # virtual host style
+    # https://bucket-name.s3-Region.amazonaws.com/key-name
+    match = re.search(r'^https?://(.+)\.s3\.([^.]+)\.amazonaws\.com/(.+)', url)
```
LaikaN57, line 1250: 1. Make the regex pattern a global so we don't have to re-compile it.
2. Use non-greedy matching for bucket (avoids bad key names) and use named groups (easier to see what we use after we match) like: `^https?://(?P<bucket>.+?)\.s3\.(?P<region>[^.]+)\.amazonaws\.com/(?P<key>.+)$`

kingpin/actors/aws/cloudformation.py
```diff
@@ -1230,3 +1238,24 @@ def _execute(self):
...
+    if match:
+        return match.group(1), match.group(2), match.group(3)
```
LaikaN57, line 1252: Use named groups.

kingpin/actors/aws/cloudformation.py
```diff
@@ -1230,3 +1238,24 @@ def _execute(self):
...
+
+    # path style request
+    # http://s3.region.amazonaws.com/bucket-name/key-name
+    regex = r'^https?://s3\.([^.]+)\.amazonaws\.com/([^\/]+)/(.+)'
```
LaikaN57, line 1256: `^https?://s3\.(?P<region>[^.]+)\.amazonaws\.com/(?P<bucket>[^/]+)/(?P<key>.+)$`

kingpin/actors/aws/cloudformation.py
```diff
@@ -1230,3 +1238,24 @@ def _execute(self):
...
+    match = re.search(regex, url)
+    if match:
+        return match.group(2), match.group(1), match.group(3)
+
+    raise Exception('Cannot parse url')
```
LaikaN57, line 1261: Maybe add native s3 style as well? `^s3://(?P<bucket>[^/]+)/(?P<key>.+)$`

mrterry:
@LaikaN57 Thanks for the review!  I've take you suggestions into https://github.com/Nextdoor/kingpin/pull/508 which is the PR that is going to move forward.

PR #508, opened 2020-08-21
Add support for CFN templates in S3

mrterry:
This implementation requires the user to explicitly provide a S3 bucket and key to their template.

### DIFF
--- a/kingpin/actors/aws/cloudformation.py
+++ b/kingpin/actors/aws/cloudformation.py
@@ -17,9 +17,11 @@
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 """
 
-import logging
 import json
+import logging
+import re
 import uuid
+from typing import Any, Dict, Optional
 
 from botocore.exceptions import ClientError
 from tornado import concurrent
@@ -28,10 +30,10 @@ from tornado import ioloop
 
 from kingpin import utils
 from kingpin.actors import exceptions
-from kingpin.actors.utils import dry
 from kingpin.actors.aws import base
-from kingpin.constants import SchemaCompareBase, StringCompareBase
+from kingpin.actors.utils import dry
 from kingpin.constants import REQUIRED, STATE
+from kingpin.constants import SchemaCompareBase, StringCompareBase
 
 log = logging.getLogger(__name__)
 
@@ -212,7 +214,7 @@ class CloudFormationBaseActor(base.AWSBaseActor):
         }
         return default_params
 
-    def _get_template_body(self, inline_template, s3_bucket, s3_key, region):
+    def _get_template_body(self, inline_template: Optional[str], s3_path: Optional[str], s3_region: Optional[str]):
         """Reads in a local template file and returns the contents.
 
         If the template string supplied is a local file resource (has no
@@ -228,6 +230,12 @@ class CloudFormationBaseActor(base.AWSBaseActor):
         Raises:
             InvalidTemplate
         """
+        # You must specify exaxtly one of inline_template or s3_path.
+        if inline_template is None and s3_path is not None:
+            raise InvalidTemplate()
+        if s3_path is None and inline_template is not None:
+            raise InvalidTemplate()
+
         # if the template is provided inline, return it and don't bother with
         # s3 buckets
         if inline_template:
@@ -236,27 +244,20 @@ class CloudFormationBaseActor(base.AWSBaseActor):
             except exceptions.UnrecoverableActorFailure as e:
                 raise InvalidTemplate(e)
 
-        try:
-            # the template is on S3.  Download it so we can inspect it and
-            # construct the proper S3 url.
-            if s3_bucket is None:
-                raise Exception('s3_bucket not specified')
-            if s3_key is None:
-                raise Exception('s3_key not specified')
+        if s3_path:
+            match = re.match(r's3://([a-z0-9.-]+)/(.*)', s3_path)
+            if match:
+                bucket = match.group(1)
+                key = match.group(2)
+            else:
+                raise InvalidTemplate()
 
-            # make sure bucket is in correct region
-            resp = self.s3_conn.get_bucket_location(s3_bucket)
-            bucket_region = resp['LocationConstraint']
+            url = f'https://{bucket}.s3.amazonaws.com/{key}'
 
-            if bucket_region != region:
-                raise Exception('bucket region does not match')
-
-            resp = self.api_call(self.s3_conn.get_object, {'Bucket': s3_bucket, 'Key': s3_key})
+            # TODO: is s3_conn pinned to aregion?
+            resp = self.api_call(self.s3_conn.get_object, {'Bucket': bucket, 'Key': key})
             remote_template = resp['Body'].read()
-            url = f'/https://s3.{region}.amazonaws.com/{s3_bucket}/{s3_key}'
             return remote_template, url
-        except Exception as e:
-            raise InvalidTemplate(e)
 
     @gen.coroutine
     def _validate_template(self, body=None, url=None):
@@ -592,7 +593,7 @@ class Create(CloudFormationBaseActor):
     not exist. Does not create the stack.
     """
 
-    all_options = {
+    all_options: Dict[str, Any] = {
         'capabilities': (list, [],
                          'The list of capabilities that you want to allow '
                          'in the stack'),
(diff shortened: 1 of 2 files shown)


kingpin/actors/aws/cloudformation.py
```diff
@@ -212,35 +212,50 @@ def _discover_default_params(self, template_body):
...
-        except exceptions.UnrecoverableActorFailure as e:
+            # the template is on S3.  Download it so we can inspect it and
+            # construct the proper S3 url.
+            if s3_bucket is None:
+                raise Exception('s3_bucket not specified')
+            if s3_key is None:
+                raise Exception('s3_key not specified')
+
+            # make sure bucket is in correct region
+            resp = self.s3_conn.get_bucket_location(s3_bucket)
```
diranged, line 248: Don't do the lookup unless we're going to use it as a backup. I suggest something along these lines:

```python
if bucket_region is None:
    try:
        resp = self.s3_conn.get_bucket_location(s3_bucket)
        bucket_location = resp["LocationConstraint"]
    except KeyError:
        raise ... "couldnt get bucket location from AWS: {resp}"
    except ClientError as e:
        raise ... "couldnt get bucket location: {e}"
```

Basically if the user supplies the region - use it.. and let the failure happen downstream if they submitted the wrong region. Otherwise look up the bucket region dynamically and use that value.

kingpin/actors/aws/cloudformation.py
```diff
@@ -212,35 +212,50 @@ def _discover_default_params(self, template_body):
...
+            bucket_region = resp['LocationConstraint']
+
+            if bucket_region != region:
+                raise Exception('bucket region does not match')
+
+            resp = self.api_call(self.s3_conn.get_object, {'Bucket': s3_bucket, 'Key': s3_key})
+            remote_template = resp['Body'].read()
+            url = f'/https://s3.{region}.amazonaws.com/{s3_bucket}/{s3_key}'
```
diranged, line 256: Whats the extra slash at the beginning? Typo?

kingpin/actors/aws/cloudformation.py
diranged: Side note.. you may get None in this value too if the bucket is in Virginia..

kingpin/actors/aws/cloudformation.py
```diff
@@ -803,9 +821,11 @@ class Stack(CloudFormationBaseActor):
...
         'role_arn': (str, None,
                      'The Amazon IAM Role to use when executing the stack'),
-        'template': (str, REQUIRED,
-                     'Path to the AWS CloudFormation File. http(s)://, '
-                     'file:///, absolute or relative file paths.'),
+        'template': (str, None,
+                     'Path to the AWS CloudFormation File. (ie file:///), '
+                     'absolute or relative file paths.'),
+        'template_bucket': (str, None, 'S3 bucket containing the CFN template'),
+        'template_key': (str, None, 'Key in S3 bucket containing the CFN template'),
```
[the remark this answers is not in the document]
diranged: So after reviewing #507 as well, I think that we should split the difference. I prefer the model now of one parameter for the stack location (`template`) - and let that parameter take two value forms:

```yaml
template: my_stack.yaml
```

or
```yaml
template: s3://bucketname/key
```

The __init__ method should then handle determining whether or not its an inline template or its an S3-hosted template.. and from there it can go and validate the template appropriately. The rest of the code in this PR i like more (rather than asking the user to construct S3 urls). We can add a `tempate_region` optional parameter if the user wants to be explicit about where the bucket is, otherwise we do the get_bucket_location lookup for them.

kingpin/actors/aws/cloudformation.py
```diff
@@ -824,11 +842,14 @@ def __init__(self, *args, **kwargs):
         # Check if the supplied CF template is a local file. If it is, read it
         # into memory.
         self._template_body, self._template_url = self._get_template_body(
-            self.option('template'))
+            self.option('template'),
+            self.option('s3_bucket'),
+            self.option('s3_key'),
+            self.option('region')
```
diranged, line 848: This enforces a requirement that the S3 bucket would _have_ to be in the same region as your target stack because you are re-using the `region` parameter to the actor. There should be a separate (optional) key for `template_region` or `template_bucket_region`.

kingpin/actors/aws/cloudformation.py
```diff
@@ -824,11 +842,14 @@ def __init__(self, *args, **kwargs):
         # Check if the supplied CF template is a local file. If it is, read it
```
diranged, line 842: I think we should raise an exception if `template` and `template_url` are supplied.. rather than just defaulting to one or the other silently.

kingpin/actors/aws/cloudformation.py
```diff
@@ -212,35 +212,50 @@ def _discover_default_params(self, template_body):
...
+                return json.dumps(self._parse_policy_json(inline_template)), None
+            except exceptions.UnrecoverableActorFailure as e:
+                raise InvalidTemplate(e)
 
         try:
-            return json.dumps(self._parse_policy_json(template)), None
-        except exceptions.UnrecoverableActorFailure as e:
+            # the template is on S3.  Download it so we can inspect it and
+            # construct the proper S3 url.
+            if s3_bucket is None:
```
diranged, line 242: Per my comment in the `__init__` method, I suggest we just do the validation at instantiation time and throw an exception if both values are supplied. We'll have to make `template` optional anyways at that point (you wouldn't want to force the user to submit some bogus string).

kingpin/actors/aws/cloudformation.py
mrterry: typo

kingpin/actors/aws/cloudformation.py
```diff
@@ -212,36 +218,63 @@ def _discover_default_params(self, template_body):
...
+                log.info(f'Getting region for bucket {bucket}')
+                resp = self.s3_conn.get_bucket_location(Bucket=bucket)
+                s3_region = resp['LocationConstraint']
+                if s3_region is None:
+                    s3_region = 'us-east-1'
+            # AWS support assured me this format would work for all regions for
+            # all buckets they also said
+            # f'https://{bucket}.s3-{region}.amazonaws.com/{key} would also
+            # work but I'm hesitant to use buckets as components in a subdomain
+            url = f'https://s3-{s3_region}.amazonaws.com/{bucket}/{key}'
```
diranged, line 259: :+1:

kingpin/actors/aws/cloudformation.py
```diff
@@ -212,36 +218,63 @@ def _discover_default_params(self, template_body):
...
+
+            s3 = self.get_s3_client(s3_region)
+            try:
+                log.info('Downloading template stored in s3')
```
diranged, line 263: Lets make this a debug statement..

kingpin/actors/aws/cloudformation.py
```diff
@@ -212,36 +218,63 @@ def _discover_default_params(self, template_body):
...
+            match = S3_REGEX.match(template)
+            if match:
+                bucket = match.group('bucket')
+                key = match.group('key')
+            else:
+                raise InvalidTemplate()
+
+            # figure out the region the bucket is in
+            if s3_region is None:
+                log.info(f'Getting region for bucket {bucket}')
```
diranged, line 250: Lets make this a debug statement

kingpin/actors/aws/cloudformation.py
```diff
@@ -212,36 +218,63 @@ def _discover_default_params(self, template_body):
...
+            # all buckets they also said
+            # f'https://{bucket}.s3-{region}.amazonaws.com/{key} would also
+            # work but I'm hesitant to use buckets as components in a subdomain
+            url = f'https://s3-{s3_region}.amazonaws.com/{bucket}/{key}'
+
+            s3 = self.get_s3_client(s3_region)
+            try:
+                log.info('Downloading template stored in s3')
+                resp = s3.get_object(Bucket=bucket, Key=key)
+            except Exception as e:
```
diranged, line 265: `ClientError`.. not `Exception` please

kingpin/actors/aws/cloudformation.py
diranged: Move this up and out of the `try` block

kingpin/actors/aws/cloudformation.py
```diff
@@ -212,36 +218,63 @@ def _discover_default_params(self, template_body):
...
+            try:
+                return json.dumps(self._parse_policy_json(template)), None
+            except exceptions.UnrecoverableActorFailure as e:
+                raise InvalidTemplate(e)
 
-        try:
-            return json.dumps(self._parse_policy_json(template)), None
-        except exceptions.UnrecoverableActorFailure as e:
-            raise InvalidTemplate(e)
+    def get_s3_client(self, region):
```
diranged, line 276: Add a docstring that identifies why we have this function at all - why you would use this instead of the built-in s3 client object for the actor..

kingpin/actors/aws/cloudformation.py
```diff
@@ -256,17 +289,16 @@ def _validate_template(self, body=None, url=None):
...
         """
 
-        if body is not None:
-            cfg = {'TemplateBody': body}
-            self.log.info('Validating template with AWS...')
+        if url is not None:
+            cfg = {'TemplateURL': url}
+            self.log.info('Validating template (%s) with AWS...' % url)
             try:
                 yield self.api_call(self.cf3_conn.validate_template, **cfg)
```
diranged, line 296: Might as well condense lines 295-298 and 302-304 into one block..

kingpin/actors/aws/cloudformation.py
```diff
@@ -804,8 +838,10 @@ class Stack(CloudFormationBaseActor):
         'role_arn': (str, None,
                      'The Amazon IAM Role to use when executing the stack'),
         'template': (str, REQUIRED,
-                     'Path to the AWS CloudFormation File. http(s)://, '
+                     'Path to the AWS CloudFormation File. s3://, '
```
diranged, line 841: Lets raise an exception if both `template` and `template_url` are supplied - we don't want ambiguous behavior here..

diranged (approved):

PR #512, opened 2021-06-14
Add support for !GetAtt, !Sub and datetime by using Amazon's cfn-flip to parse YAML

corydolphin:
While trying to add Datadog's [MetricStream CFN stack](https://datadog-cloudformation-stream-template.s3.amazonaws.com/aws/streams_main.yaml), I hit two issues with KingPin.

1. It uses the shorthand functions `!GetAtt` and `!Sub` a lot. 
2. The `Version` in Policy documents was unquoted, so PyYaml natively parsed it as a date which then failed to JSON dump. 

This is currently failing tests due to coveragee, but I do not see many examples of YAML based CFN stacks that would be easy to build upon. 

@diranged @stlava Does this seem like a reasonable idea to you two?

### DIFF
--- a/examples/test/aws.cloudformation/cf.integration.yaml
+++ b/examples/test/aws.cloudformation/cf.integration.yaml
@@ -1,0 +1,30 @@
+---
+AWSTemplateFormatVersion: 2010-09-09
+Description: Stupid-simple CloudFormation template for Kingpin Integration Testing.
+Parameters:
+  BucketName:
+    Type: String
+  BucketPassword:
+    Type: String
+    NoEcho: true
+  DefaultParam:
+    Type: String
+    Default: DefaultValue
+  Metadata:
+    Type: String
+Conditions: {}
+Resources:
+  IntegrationTestBucket:
+    Type: AWS::S3::Bucket
+    Metadata:
+      Test:
+        Ref: Metadata
+    Properties:
+      BucketName: !Ref BucketName
+      Tags:
+      - Key: DefaultParam
+        Value:
+          Ref: DefaultParam
+      - Key: Kingpin
+        Value:
+          Ref: BucketName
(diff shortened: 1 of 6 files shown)


kingpin/actors/aws/cloudformation.py
```diff
@@ -40,6 +43,11 @@
 
 __author__ = 'Matt Wise <matt@nextdoor.com>'
 
+class DateTimeEncoder(JSONEncoder):
+    #Override the default method
+    def default(self, obj):
+        if isinstance(obj, (datetime.date, datetime.datetime)):
+            return obj.isoformat()
```
sabw8217, line 50: isn't this going to turn something like `2010-12-10` into `2010-12-10T00:00:00`?

I think `_parse_policy_json` would need to not parse `2010-12-10` as a datetime in the first place but that seems like a potentially much more intrusive change.

corydolphin: Let me actually remove the `datetime.datetime` support. These end up as `datetime.date`.

```
import datetime
datetime.date.fromisoformat('2010-12-10').isoformat()
'2010-12-10'
```

But, I am definitely a little worried about this. In the Boneman change, would a change in a policy version show up in the diff?

diranged: So AWS hasn't changed the "version"  since 2010.. so just FWIW, it's not likely that we will be changing that string any time soon.

examples/test/aws.cloudformation/cf.integration.yaml
```diff
@@ -0,0 +1,30 @@
+---
```
diranged, line 1: Can you add a test CFN stack with the `!` shortcuts to prove it works?

diranged (approved):
Looks good to me!

PR #521, opened 2022-09-02
Update boto3 client with aws_session_token

davidwhite-nd:
Updating the boto3 client create to add aws_session_token as a parameter. 

Was getting the following error when trying to validate a template in a dryrun:
An error occurred (InvalidClientTokenId) when calling the ValidateTemplate operation: The security token included in the request is invalid.

After adding the aws_session_token to the boto3 client, kingpin was able to validate the template.


davidwhite-nd:
testing integration

LaikaN57:
I am going to dig into this a bit before reviewing. Please stand by...

kingpin/actors/aws/test/integration_s3.py
```diff
@@ -27,8 +27,8 @@ class IntegrationS3(testing.AsyncTestCase):
     Requirements:
         You have to create an S3 Bucket named kingpin-integration-test and
         place it in the specified region (default us-east-1).
-        As with other tests, environment variables AWS_ACCESS_KEY_ID and
-        AWS_SECRET_ACCESS_KEY are expected, and the key should have
+        As with other tests, environment variables AWS_ACCESS_KEY_ID,
+        AWS_SECRET_ACCESS_KEY and AWS_SESSION_TOKEN are expected, and the key should have
```
stlava, line 31: Nit missing comma.

kingpin/actors/aws/test/integration_cloudformation.py
```diff
@@ -103,8 +104,9 @@ class IntegrationStack(testing.AsyncTestCase):
     * Delete that same stack
 
     Requirements:
-        Your AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY must have access to
-        create CF stacks. The stack we create is extremely simple, and should
+        Your must have an AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and AWS_SESSION_TOKEN
```
stlava, line 107: Nit should say something like: access key id and access key and optionally session token when using an assumed role.

corydolphin: Besides these comments, is there anything else blocking?

davidwhite-nd: changed the verbiage to:

        You must have an AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY if using
        an IAM user/role and also an AWS_SESSION_TOKEN if using temporary access credentials,
        with permissions to ______

davidwhite-nd:
Updated comments, please let me know if there's any other updates needed

kingpin/actors/aws/test/integration_s3.py
davidwhite-nd: updated verbiage

        You must have an AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY if using
        an IAM user/role and also an AWS_SESSION_TOKEN if using temporary access credentials,
        with permissions to

kingpin/actors/aws/base.py
```diff
@@ -86,19 +90,23 @@ def __init__(self, *args, **kwargs):
...
-            secret = aws_settings.AWS_SECRET_ACCESS_KEY
+        #
+        # Note: these get defualted to None in aws_settings if they are not
+        # found which will tell boto3 to fallback to default behavior.
+        boto3_client_kwargs = {}
+        boto3_client_kwargs["aws_access_key_id"] = aws_settings.AWS_ACCESS_KEY_ID
+        boto3_client_kwargs[
+            "aws_secret_access_key"
+        ] = aws_settings.AWS_SECRET_ACCESS_KEY
+        boto3_client_kwargs["aws_session_token"] = aws_settings.AWS_SESSION_TOKEN
```
LaikaN57, line 105: @davidwhite-nd So it looks like this might only be valuable for unit testing by the looks of it. Other than that, it looks like we can completely remove this code.

But for now, I am just going to merge this PR. Alternatively, you can just unset the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` env vars before calling kingpin to get the same effect on older versions (as long as your keys are stored in a creds file.

Also, can you reverse-review this. I believe it is safe to pass in an unchecked dict as kwargs as the underlying boto3 library seems to handle None values gracefully (ie. reverts back to the default searching).

LaikaN57 (approved):

kingpin/actors/aws/base.py
```diff
@@ -86,19 +90,23 @@ def __init__(self, *args, **kwargs):
...
         # keys in the environment (either for the purposes of a unit test, or
         # because they wanted to), we use those values.
-        if aws_settings.AWS_ACCESS_KEY_ID and aws_settings.AWS_SECRET_ACCESS_KEY:
-            key = aws_settings.AWS_ACCESS_KEY_ID
-            secret = aws_settings.AWS_SECRET_ACCESS_KEY
+        #
+        # Note: these get defualted to None in aws_settings if they are not
+        # found which will tell boto3 to fallback to default behavior.
+        boto3_client_kwargs = {}
+        boto3_client_kwargs["aws_access_key_id"] = aws_settings.AWS_ACCESS_KEY_ID
```
corydolphin, line 101: General feedback: separating structural changes and behavioral changes will make this easier to read. Eg first the change to move to kwargs.

kingpin/actors/aws/base.py
```diff
@@ -86,19 +90,23 @@ def __init__(self, *args, **kwargs):
...
+        boto3_client_kwargs[
+            "aws_secret_access_key"
+        ] = aws_settings.AWS_SECRET_ACCESS_KEY
```
corydolphin, line 104: What is the expected behavior if all three are set? Should we enforce an invariant?

kingpin/actors/aws/base.py
```diff
@@ -86,19 +90,23 @@ def __init__(self, *args, **kwargs):
...
+        boto3_client_kwargs["aws_session_token"] = aws_settings.AWS_SESSION_TOKEN
 
         # Establish connection objects that don't require a region
         self.iam_conn = boto3.client(
-            "iam", aws_access_key_id=key, aws_secret_access_key=secret
+            service_name="iam", config=None, **boto3_client_kwargs
         )
```
corydolphin, line 110: 1. I had assumed the use of the kwargs was to ensure we don’t pass none values. What is the motivation of using a scattered kwargs dictionary here? 
2. What is the motivation for the change from positional to kwargs and explicitly setting config=None? Is that critical? It looks to only be done in this case.

kingpin/actors/aws/settings.py
```diff
@@ -31,3 +31,4 @@
 # During tests, we mock these out to blank strings to prevent these calls.
 AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY_ID", None)
 AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY", None)
+AWS_SESSION_TOKEN = os.getenv("AWS_SESSION_TOKEN", None)
```
corydolphin, line 34: This may be obvious, but I’m new to this area of the code and trying to understand. From a design point of view, if we didn’t parse these in settings, and depended on boto3’s handling, would that work? It feels like we are duplicating logic already in the SDK?

corydolphin (approved):
Left non blocking feedback. This looks reasonable to me, is there anything else blocking merging?

kingpin/actors/aws/base.py
LaikaN57: If all 3 are set, all 3 should be passed to `boto3.client()`. We will let boto3 handle error conditions between these args.

kingpin/actors/aws/base.py
LaikaN57: So for no. 1, we are actually doing the opposite. We want to be able to easily pass `None` values. For no. 2, I am moving away from positional args here just to make it safer than mixing positional/named/kwargs. (And yes, to specifically to set `config=None` so that it just matches the calls below.

kingpin/actors/aws/settings.py
LaikaN57: Yes, exactly what my approval comment pointed out as well. However, I was not 100% sure so the safe thing to do is just do the same env var override that we were doing before, but just capture this missed env var (which lead to some rewrite of the `if` statement in the beginning of the `aws/base.py` code.)

kingpin/actors/aws/base.py
LaikaN57: You bring up a good point in number 1 tho. Let me quickly test if passing a dict with None values actually passes that None to the function or not.

PR #522, opened 2022-12-22
[AWSBaseActor] Respect boto3/AWS retry configurability

LaikaN57:
We simply forward the `AWS_*` env vars into the config if they are set. Otherwise we will just use some sane default parameters which we have found work well based on our understanding of the commented documentation and our experience in running kingpin at scale.

### DIFF
--- a/kingpin/actors/aws/base.py
+++ b/kingpin/actors/aws/base.py
@@ -119,7 +119,8 @@ class AWSBaseActor(base.BaseActor):
         boto_config = botocore_config.Config(
             region_name=self.region,
             retries={
-                "mode": "adaptive",
+                "max_attempts": aws_settings.AWS_MAX_ATTEMPTS,
+                "mode": aws_settings.AWS_RETRY_MODE
             },
         )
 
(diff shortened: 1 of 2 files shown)


RKTMN (approved):

kingpin/actors/aws/base.py
```diff
@@ -119,7 +119,8 @@ def __init__(self, *args, **kwargs):
         boto_config = botocore_config.Config(
             region_name=self.region,
             retries={
-                "mode": "adaptive",
+                "max_attempts": aws_settings.AWS_MAX_ATTEMPTS,
+                "mode": aws_settings.AWS_RETRY_MODE,
```
corydolphin, line 123: It seems like we are reimplementing Boto’s environment variable usage. Would it be cleaner if we removed kingpin from that entirely by removing any explicit override?
We could document the behavior.

LaikaN57: I was attempting to maintain some "optimized sane defaults" in the case where the user does not specify anything. Because kingpin can be high call volume, the defaults for boto3 may not be suitable OOB.

PR #524, opened 2023-01-04
[ci/cd] Migrated from `setup.py` to `setuptools`/`build`

LaikaN57:
Migrating from our old `setup.py` to `setuptools` unfortunately requires a bit of a mass overhaul of the testing frameworks we used.

I will attempt to provide a summary of changes here:

- Integration test frameworks were removed as they were unused. I left the actual tests in the directories so people can still use them if they want/need
- Migrated from `nose`/`converage` to `pytest`/`pytest-cov`
- Split `lint`, `test`, and `build` into separate `make` recipes
- Removed old informational docs now found in Github history
- Fix some warnings on `test_group` related to the data classes accidentally having the prefix "Test"
- Fix some errors on `test_base` related to the module needing to be reloaded for a particular test which checks an env var was set on module load
- Removed `pep8` in favor of `black`

## Testing

I have tested all the commands in the Makefile and they worked.


LaikaN57:
See similar implementations that have been applied here: https://github.com/Nextdoor/tornado_rest_client

.circleci/config.yml
```diff
@@ -16,8 +16,9 @@ jobs:
       - restore_cache:
           keys:
             - pip-{{ checksum "requirements.test.txt" }}-2023-01-03
+      - run: make lint
```
RKTMN, line 19: Note to self: Pyblack, etc... is done in `make lint`, now

requirements.txt
```diff
@@ -15,15 +12,17 @@ cfn-flip
...
 # Used to turn snake_case into CamelCase
 inflection
 zipp
 configparser
+
+#
+# ND libraries
+#
+
+# tornado rest client
```
RKTMN, line 27: Nit: This might go better just below the Tornado requirement listing at the top

setup.cfg
```diff
@@ -0,0 +1,28 @@
...
+    Programming Language :: Python :: 3
+    Topic :: Software Development
+license = Apache License, Version 2.0
+description = Deployment Automation Engine
+long_description = file: README.rst
+keywords = deployment, automation, kingpin, aws
+
+[options]
+install_requires = file: requirements.txt
+python_requires = ~=3.7
```
RKTMN, line 22: For the future: I believe Python 3.6 recently lost support, so it may eventually be a project to get this up to Python 3.11 or whatever. Until then, thanks for undertaking this effort in general!

RKTMN (approved):
Overwhelmingly positive changes, thanks for leaving a clear description of everything that changed!

PR #537, opened 2025-01-13
feat(cloudformation): Add env var for default CFN role ARN

LaikaN57:
### Description

Mainly

- Added a new env var `KINGPIN_CFN_DEFAULT_ROLE`

Supplimetry

- Docfixes and removal of per-file lincense text in favor of repo license text

### Motivation

We wanted give our build roles tight privledges on only interact directly with the AWS API for things that `kingpin` can manage directly (ie. s3, iam, cfn). However, we want to pass another role to the CFN service to use to create all kinds of resources that `kingpin` does not natively manage.


kingpin/actors/aws/cloudformation.py
```diff
@@ -563,6 +551,8 @@ def _create_stack(self, stack):
 
         if self.option("role_arn"):
             cfg["RoleARN"] = self.option("role_arn")
+        elif KINGPIN_CFN_DEFAULT_ROLE_ARN:
+            cfg["RoleARN"] = KINGPIN_CFN_DEFAULT_ROLE_ARN
```
LaikaN57, line 555: This is the only "real" change in this PR. There is one other place we do this setting and there is an entry in the aws settings file to support this as an env var.

PR #545, opened 2025-01-21
fix(ci): Remove private GHA mechanisms

LaikaN57:


diranged (approved):
looks basically right

PR #629, opened 2025-11-03
feat(group): remove deprecated dict format for contexts option

LaikaN57:
## Summary

This PR removes the deprecated dict format for the `contexts` option in group actors and bumps the version to 5.0.0 to indicate a breaking change.

## Breaking Change

The `contexts` option in `group.Sync` and `group.Async` actors no longer accepts the deprecated dict format `{"file": "path", "tokens": {...}}`. This format was marked for deprecation in v0.5.0.

### Migration Guide

**Before (deprecated):**
```json
{
  "contexts": {
    "file": "contexts.json",
    "tokens": {"KEY": "value"}
  }
}
```

**After:**
```json
{
  "contexts": "contexts.json",
  "init_tokens": {"KEY": "value"}
}
```

## Changes

- ✅ Removed deprecated dict validation in `BaseGroupActor.__init__`
- ✅ Removed deprecated dict processing in `_build_actions` method
- ✅ Updated `all_options` to accept only `(str, list)` types
- ✅ Updated docstrings for `Sync` and `Async` actors to remove deprecated format references
- ✅ Removed tests for deprecated functionality (`test_build_actions_with_bad_context_file`, `test_build_actions_with_context_file`)
- ✅ Bumped version from 4.0.2 to 5.0.0

## Supported Formats

The `contexts` option now only accepts:
- **`list`** - A list of dictionaries
- **`str`** - A file path to a JSON/YAML file

## Test Plan

- [ ] Run existing unit tests to ensure no regressions
- [ ] Verify that string format for contexts still works correctly
- [ ] Verify that list format for contexts still works correctly
- [ ] Confirm deprecated dict format is no longer accepted

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kingpin/actors/aws/base.py
+++ b/kingpin/actors/aws/base.py
@@ -171,33 +171,34 @@ class AWSBaseActor(base.BaseActor):
             return exceptions.RecoverableActorFailure("Boto3 had a failure: %s" % e)
         return e
 
-    def _parse_policy_json(self, policy):
-        """Parse a single JSON file into an Amazon policy.
+    def _parse_json(self, file_path):
+        """Parse a single JSON file.
 
-        Validates that the policy document can be parsed, strips out any
+        Validates that the JSON document can be parsed, strips out any
         comments, and fills in any environmental tokens. Returns a dictionary of
         the contents.
 
         Args:
-            policy: The Policy JSON file to read.
+            file_path: The JSON file to read.
 
         Returns:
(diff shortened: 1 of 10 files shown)


LaikaN57:
Motivation for change...

This is some precursor work to actually re-implement allowing map types but with a different shape and for a different purpose - mergable contexts.

We could keep both things but it is going to be easier to develop if we remove the existing deprecated map style/shape first and avoid the overload.

